PR #181, opened 2015-07-01
Update related to Speed branch from scheduling

lpellegr:


lpellegr:
@youribonnaffe Could you please have a look?

youribonnaffe:
When adding a environment variable (without setting appendChar value), checking of the workflow will fail (_At [task=Linux] [tag:job] : Error: attribute "appendChar" has a bad value: the value does not match the regular expression ".". at line 16 , column 77._)
One more point to remove it :)

youribonnaffe:
Hitting the delete key with the focus in a text field (for instance the command to run) deletes the task

youribonnaffe:
I think I found the issue (see d994c7d).
The keyboard handling was just there inside the JobClasspath function...

lpellegr:
:100:

PR #220, opened 2016-02-12
Prevent the Studio test run into an infinite loop of failures.

tobwiens:
Prevent the Studio test run into an infinite loop of failures.

The selenium tests have some parts, which retry in an infinite loop. That can and did let the test run into an infinite loop, blocking jenkins and flooding the file system with error reports. To prevent that from happening again, a maximum number of tries is introduced.

### DIFF
--- a/test/ui/commands/freshWorkflow.js
+++ b/test/ui/commands/freshWorkflow.js
@@ -4,16 +4,17 @@
 exports.command = function () {
     var browser = this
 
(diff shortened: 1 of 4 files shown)


tobwiens:
@ow2-proactive/developers: Can you have a look. I checked it locally. Works better now, still not perfect though.

paraita:
:+1:

PR #240, opened 2016-03-18
Change padding, size of the tool buttons and add icons

ShatalovYaroslav:
Change padding, size of the tool buttons and add icons


marcocast:
Looks good to me!

Issue #245, opened 2016-04-05
Workflow Title is not displayed


tobwiens:
![bug1missingworkflowname](https://cloud.githubusercontent.com/assets/6491363/14275097/938ba002-fb15-11e5-9650-7fa4d0b17eff.png)

lpellegr:
Should be fixed by 99937e666f3b9fc654cdb6917170ee15d773813f.

tobwiens:
Fixed by 99937e666f3b9fc654cdb6917170ee15d773813f

PR #247, opened 2016-04-05
Fix failing "Pull out a dependency" Studio test

tobwiens:
The studio test parses the xml in the export window. The parsing method looks for the jobdescriptor 3.3. We updated the jobdescriptor to 3.4, so this test needs to be adjusted as well.


tobwiens:
I merged the other PR, https://github.com/ow2-proactive/studio/issues/245, accidentally in the pause-on-task-error-branch

PR #250, opened 2016-04-20
Added support for modified script tasks

marcocast:


app/scripts/proactive/view/TaskView.js
```diff
@@ -37,7 +37,11 @@ define(
             var iconPath = this.icons[this.modelType];
             try{
 	            iconPath = this.iconsPerLanguage[this.model.get("Execute").get("Script").get("Language")];
-            }catch(err){}
+            }catch(err){
+            	 try{
+     	            iconPath = this.iconsPerLanguage[this.model.get("Execute").Script.Language];
```
tobwiens: Why do we have those two different ways to access it. 

I guess, that when we initially create the model we create it differently compared to when we change the script. 
It seems that we solve the symptom but that is not really the issue.

marcocast: Yes this is a workaround. maybe I ll spend  some more time trying to understand the whys and the hows.

paraita: it is possible to access json attributes in javascript either by their attribute key as a string, or directly accessing it with the dot notation.

tobwiens: I looked into it a little bit:
the .get() is from backbonejs http://backbonejs.org/#Model-get
Whereas . and [] afre the standard javascript access: http://www.w3schools.com/js/js_objects.asp

I don't see any piece of code which changes the script object:

Here it is initialized 

```
var script = new Script();
                script.set("Script", "print(java.lang.System.getProperty('pas.task.name'))");
                script.set("Language", "javascript");
                this.model.get("Execute").set("Script", script);
```
And here it is updated: 

```
changeTaskType: function () {
            var executableTypeStr = this.model.get("Type");
            
            if (this.modelType && this.modelType != executableTypeStr) {
                var executableType = require('proactive/model/'+executableTypeStr);
                var executable = new executableType();

                this.model.schema = $.extend(true, {}, this.model.schema);
                // TODO beautify
                if (executableTypeStr == "JavaExecutable") {
                    this.model.schema['Execute'] = {type: 'NestedModel', model: JavaExecutable};
                } else if (executableTypeStr == "NativeExecutable") {
                    this.model.schema['Execute'] = {type: 'NestedModel', model: NativeExecutable};
                } else {
                    this.model.schema['Execute'] = {type: 'NestedModel', model: ScriptExecutable};
                    executable["Script"] = {"Language": "bash"};
                }
                this.$el.find("img").attr('src', this.icons[executableTypeStr]);
                this.model.set({"Execute": executable});
                this.$el.click();
            }
            this.modelType = executableTypeStr;
        }
```

There is the change in the model:

```
executable["Script"] = {"Language": "bash"};
```

marcocast: This line

`executable["Script"] = {"Language": "bash"};`

is not executed when the model changes.

PR #252, opened 2016-05-02
Allow parallel environments with numberOfNodes=1

tobwiens:
Allow parallel environments with numberOfNodes=1

Parallel environments can have 1 node.


tobwiens:
@fviale This fixes the singleHostExclusive issue with one node #252 . Can you have a look over the code?

fviale:
:+1:

PR #257, opened 2016-05-17
fixes #254: Job priority is lost when a workflow is imported

### LINKED ISSUE #254: Job priority is lost when a workflow is imported
Let's say you have a workflow which has been exported from the Studio. When the workflow definition is imported, the Job priority is lost. Job priority field has "normal" value selected.

The issue occurs on try with version 7.8.0 but also on trydev with current master.

Below is the workflow I used to reproduce the issue:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<job
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xmlns="urn:proactive:jobdescriptor:3.4"
     xsi:schemaLocation="urn:proactive:jobdescriptor:3.4 http://www.activeeon.com/public_content/schemas/proactive/jobdescriptor/3.4/schedulerjob.xsd"
    name="SingleTaskHighestPriority" 
    priority="highest"
    onTaskError="continueJobExecution"
     maxNumberOfExecution="2"
>
  <description>
    <![CDATA[ A workflow that executes python in JVM. ]]>
  </description>
  <taskFlow>
    <task name="HighestPriorityTask">
      <description>
        <![CDATA[ The simplest task, ran by a python engine. ]]>
      </description>
      <scriptExecutable>
        <script>
          <code language="python">
            <![CDATA[
                    for x in range(1, 11):
    print x
                    ]]>
          </code>
        </script>
      </scriptExecutable>
    </task>
  </taskFlow>
</job>
```

sophiesongge:
By allowing the override of the existing parts.

### DIFF
--- a/app/scripts/proactive/model/SchemaModel.js
+++ b/app/scripts/proactive/model/SchemaModel.js
@@ -109,13 +109,9 @@ define(
                                     console.log("Should no be here", prop, value);
                                 }
                             } else {
-                                if (merging && that.get(prop)) {
-                                    // do not override existing value when merging, except for workflow name
-                                } else {
-//                                    console.log("Setting", prop, "from", placeholder, "to", value)
-                                    value = value.trim()
-                                    that.set(prop, value)
-                                }
+                                  console.log("Setting", prop, "from", placeholder, "to", value)
+                                                                    value = value.trim()
+                                                                    that.set(prop, value)
                             }
                         }
                     }
--- a/app/scripts/proactive/model/SchemaModel.js
+++ b/app/scripts/proactive/model/SchemaModel.js
@@ -45,6 +45,7 @@ define(
                 this.set("On Task Error Policy", "cancelJob")
             }
         },
+        //This method is used for the imported job xml
         populateSchema: function (obj, merging) {
             var that = this;
 
@@ -118,6 +119,7 @@ define(
                 }
             }
         },
+        //This method is used for the templates
         populateTemplate: function (obj, merging) {
             var that = this;
 
@@ -183,7 +185,13 @@ define(
                                 }
                             } else {
                                 if (merging && that.get(prop)) {
-                                    // do not override existing value when merging, except for workflow name
+                                    // do not override existing value when merging
+                                } else if(this.schema["Name"]){
+                                   // do not override workflow name even if it is empty
+                                } else if(this.schema["Project"]){
+                                   // do not override project name even if it is empty
+                                } else if(this.schema["Description"] ){
+                                   // do not override project description even if it is empty
                                 } else {
 //                                    console.log("Setting", prop, "from", placeholder, "to", value)
                                     value = value.trim()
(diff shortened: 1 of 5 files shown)


sophiesongge:
Before, we don't want the new imported xml file to override the already existed parts, so, every already existed field will not be changed except for workflow name. Because of this, not only the "Job Priority", but also all the other field who already has an initial value, or is set to a value manually would not be changed by the new imported xml file.

lpellegr:
@sophiesongge I have tested your patch. The behaviour looks much better. However, I noticed that if you create a new workflow, import a new one that has no _Project_ name (empty field) and finally you drag and drop  _MPI Job_ from _Templates_, then you will have the Project name set to _Advanced Workflows_. 

I understand that it happens because the field was empty and to replace something empty by another value is not considered as overriding but it looks strange to me, especially for Project name and description. It's my own opinion but I think that Workflow name, description and Project values should not be changed (even if they are empty) when we drag and drop a template.

Regarding the code, it looks like the piece of code added in commit d4e13b0e87d09008d8b3c00033dce98d3a7cacf4 is a duplicate of an existing one. It would be nice, it if possible, to factorize it by using a shared method.

sophiesongge:
retest please

sophiesongge:
Tow reviews please

sophiesongge:
Two reviews please :)

app/scripts/proactive/model/SchemaModel.js
```diff
@@ -109,13 +109,9 @@ define(
...
                             } else {
-                                if (merging && that.get(prop)) {
-                                    // do not override existing value when merging, except for workflow name
-                                } else {
-//                                    console.log("Setting", prop, "from", placeholder, "to", value)
-                                    value = value.trim()
-                                    that.set(prop, value)
-                                }
+                                  console.log("Setting", prop, "from", placeholder, "to", value)
+                                                                    value = value.trim()
```
tobwiens: This seems to be not the right formatting

app/scripts/proactive/model/SchemaModel.js
```diff
@@ -109,9 +109,9 @@ define(
                                     console.log("Should no be here", prop, value);
                                 }
                             } else {
-                                  console.log("Setting", prop, "from", placeholder, "to", value)
-                                                                    value = value.trim()
-                                                                    that.set(prop, value)
+//                                  console.log("Setting", prop, "from", placeholder, "to", value)
+                                    value = value.trim()
```
tobwiens: Can you merge those commits?

tobwiens:
:+1: I like us to find a way to increase readability inside our javascript code.

app/scripts/proactive/model/SchemaModel.js
sophiesongge: Okay, I will merge them

tobwiens:
The code is okay, you can merge it. But I like to start a discussion about increasing readability of our javascript code.

PR #259, opened 2016-05-23
Add unit test for Python formatting script

marcocast:


marcocast:
2 reviews please?

tobwiens:
:+1: Looks good. Sadly it is bound to the exact script we use.
Looks like we can easily add this for other languages as well.

fviale:
Same remark as tobias, but I don't know how to make it different

marcocast:
@tobwiens  Yes we could add a test for each script type easily, but i felt was out of the scope of this story

PR #280, opened 2016-08-04
Adding a cache mechanism to dataspace

fviale:
https://github.com/ow2-proactive/scheduling/issues/2619
- added new transfer directives in studio
- changed job schemas to version 3.5


paraita:
LGTM

PR #281, opened 2016-08-04
Fixes #278: Message for authentication failure is misleading

### LINKED ISSUE #278: Message for authentication failure is misleading
When wrong login and/or password are used, the following message appears:

capture d ecran de 2016-07-06 11-43-34

It is misleading. We are receiving support message due to that because the user does not understand what happen. It would be much better to display something like "Invalid login or password" but also to differentiate an issue because the REST API is not reachable from an authentication failure due to login or password.

sophiesongge:
By adding judgement about invalid login and password


paraita:
LGTM

fviale:
:+1:

PR #282, opened 2016-08-08
Create a new menu for notification services

sophiesongge:
create a new menu for notification services


cdiop:
LGTM

lpellegr:
Images which have been added should be resized.

@sophiesongge If I understand, Michael and I will have to remove our workflows from the Scheduling projects so that they no longer appear in the _templates_ menu? instead you will include them in the Studio directly?

lpellegr:
I haven't followed the discussions since I was not there but do you know what is the reason to name the new menu _Manuals_?

sophiesongge:
@lpellegr
Thanks.
For the templates, I can find a way to import it from scheduling, but I don't have enough time for this release. So, I think it is better to keep it in Scheduler for now.

sophiesongge:
@lpellegr There is an email talking about this menu, I will send it to you.

PR #297, opened 2016-08-31
New WebValidation task in Manual tasks

marcocast:
Python task that pauses its job and send a  notification validation message to the notification-service


sophiesongge:
I think you need also to add an image in the images folder with the name: WebValidationNotification.png , then it will be okay

mbenguig:
OK

PR #300, opened 2016-09-06
Change padding size due to su menus

tobwiens:
- Sub menus provide more space in the navigation bar, revert the previously added padding

@sophiesongge Thank you for pointing out that we changed the padding recently so that the "big" menu fits. 

This PR reverses the padding change, because the sub menus reduce the menu size significantly.

Here is the Studio with padding: 
image

Here is the new version without padding:
image

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -29,7 +29,7 @@
         <!-- endbuild -->
 
     </head>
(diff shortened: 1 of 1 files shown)


mbenguig:
Ok

PR #301, opened 2016-09-07
Improve test documentation

tobwiens:
Improve test documentation


tobwiens:
I took your explanations @paraita and made the test documentation out of it, thank you for it.

Are there more chapters one like to see?

Issue #312, opened 2016-10-20
Login sometimes requires 2 submissions


lpellegr:
Studio portal often requires to authenticate twice.

Below are steps to reproduce:

1. Close your browser if it is open.
2. Stop any Scheduler instance.
3. Start a Scheduler instance.
4. Open your browser with 2 tabs, move to the second and open http://localhost:8080/studio.
5. Use admin/admin as credentials and submit form (at this stage we don't care if this first authentication has required double submission or not).
6. Close your second tab.
7. Restart your Scheduler instance.
8. Open http://localhost:8080/studio in a new tab
9. Submit the form. At this step the authentication should not work.
10. Submit again the form and you are authenticated.

PR #313, opened 2016-10-24
Append template content to document object at first during rendering

lpellegr:
Template content associated to login view was not always appended to the
document object while rendering the login view with Backbone.js. Also, binding
was too late for some of the cases that were already handled. As a consequence,
the event that is defined for the Submit button was not registered by
Backbone.js and thus not triggered.

This patch fixes ow2-proactive/studio#312.


lpellegr:
@ow2-proactive/developers Could 2 people please review the changes?

PR #315, opened 2016-11-10
Task variables

laurianed:
A panel is displayed to enable the user to add task variables.
Variables are displayed into text areas so that the user is aware that he can update them by clicking on them.


sophiesongge:
LGTM

PR #316, opened 2016-11-16
Task variables studio improvement

laurianed:
Correct missing XML task variables part
Studio variables improvement

### DIFF
--- a/app/scripts/proactive/model/utils.js
+++ b/app/scripts/proactive/model/utils.js
@@ -1,5 +1,5 @@
 define(function (require) {
-
+	
     return {
         inlineNameNotEditable: function(prop) {
             var name = prop['Name'] ? prop['Name'] : prop['Property Name'];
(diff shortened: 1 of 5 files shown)


app/scripts/proactive/model/utils.js
```diff
@@ -18,13 +18,17 @@ define(function (require) {
         },
         inlineNameValueInherited: function(prop) {
         	var checked = "";
+        	// update datatype if wrong, otherwise dislay is not correct
+        	if (prop['Inherited'] == "true") prop['Inherited'] = true;
+        	if (prop['Inherited'] == "false") prop['Inherited'] = false;
+
         	if (prop['Inherited']){
         		checked=" checked"
         	}
```
sophiesongge: Here, can we use if(prop['Inherited']=="true"), instead of change it, if we don't need to use it later as a boolean? Then we can save 2 lines.

PR #335, opened 2017-01-27
Add a new button in manuals for email validation

laurianed:


app/templates/email_validation.xml
```diff
@@ -0,0 +1,60 @@
...
+      </description>
+      <scriptExecutable>
+        <script>
+          <code language="python">
+            <![CDATA[
+# Please fill variables
+
+notification_message = 'Put your email validation message here'
+email_to = 'user@domain.org'
+user_to_name = 'user'
```
ShatalovYaroslav: Maybe these variables can be set up as task variables, as it can be more user-friendly

PR #341, opened 2017-02-08
Workflow catalog UI improvement for ZIP import and export

laurianed:


laurianed:
https://github.com/ow2-proactive/workflow-catalog/pull/27

PR #342, opened 2017-02-08
Ldap

vinseon:
Add LDAP variables: `ldapSearchBase`, `ldapUsername` and update given examples


vinseon:
The check fail is due to the previous pull request: https://github.com/ow2-proactive/studio/pull/341

PR #347, opened 2017-02-17
Workflow Catalog UI minor modifications

laurianed:


app/index.html
```diff
@@ -232,11 +232,11 @@
                     	<table style="width:100%;"><tr><td>
 		                <table><tr><td><select class="form-control" id="catalog-select-action">
 		                    <option value="-1" selected="selected">Please choose an action</option>
-		                    <option value="0" id="publish-to-remote">Send to another Scheduler</option>
-		                    <option value="1" id="export-as-archive-button">Export selected workflows</option>
-		                    <option value="2" id="import-archive-button">Import archive of workflows</option>
-		                    <option value="3" id="delete-selection-catalog">Delete selected workflows</option>
-		                    <option value="4" id="publish-to-catalog-button">Add open workflow to this bucket</option>
+		                    <option value="0" id="publish-to-remote">Send to another Workflow Catalog</option>
+		                    <option value="1" id="export-as-archive-button">Export selected Workflows</option>
```
MaelAudren: Why do you put an uppercase to "workflow" ?

laurianed: Because I think the rule is to put an uppercase for all activeeon objects (Workflow, Task)

MaelAudren: You right : https://github.com/ow2-proactive/documentation/blob/master/README.adoc

PR #350, opened 2017-02-24
Quality Check: handle the case where the first variable defined in a …

fviale:
…task or job contains a Model

- at this time, the Variables attribute is not yet defined, use the variable received from the form to define it.


fviale:
Retest this please

PR #358, opened 2017-03-27
submit cron workflow to execution scheduler service

yinan-liu:


app/scripts/proactive/config.js
```diff
@@ -6,6 +6,7 @@ define(function () {
     
     return {
         restApiUrl: '/rest/studio',
+        execution_scheduler_restApiUrl: '/execution-scheduler/scheduled-workflows',
```
tobwiens: Can we make this configurable with an IP/Hostname. The service could run on another machine.

app/scripts/proactive/rest/studio-client.js
```diff
@@ -202,16 +202,59 @@ define(
...
+                    type: 'POST',
+                    beforeSend: function (xhr) {
+                        xhr.setRequestHeader('sessionid', localStorage['pa.session'])
+                    },
+                    success: function (data) {
+                        console.log("Should not be there", data);
+                    },
+                    error: function (data) {
+                        if (data.status == 200) {
+                            console.log("Success", data);
```
tobwiens: Why is the submission success handled inside the error handler?

yinan-liu: actually i dont know, i copy paste the existing code from other fuctions

yinan-liu: i will test to see

app/scripts/proactive/config.js
yinan-liu: I agree with you. But i am not sure it could be the case for us, since the service's war is packaged along with the scheduler release package, and gets deployed in the same jetty server.

PR #359, opened 2017-03-28
Escape Generic Information

tobwiens:
Escape Generic Information

### DIFF
--- a/app/scripts/proactive/templates/job-template.html
+++ b/app/scripts/proactive/templates/job-template.html
@@ -21,7 +21,7 @@
     <% if (job["Generic Info"] && job["Generic Info"].length>0) { %>
     <genericInformation>
     <% _.each(job["Generic Info"], function(gi) { %>
-            <info name="<%=gi["Property Name"]%>" value="<%=gi["Property Value"]%>"/>
+            <info name="<%=gi["Property Name"]%>" value="<%= _.escape(gi["Property Value"]) %> "/>
     <% }) %>
     </genericInformation>
     <% } %>


app/scripts/proactive/templates/job-template.html
```diff
@@ -21,7 +21,7 @@
     <% if (job["Generic Info"] && job["Generic Info"].length>0) { %>
     <genericInformation>
     <% _.each(job["Generic Info"], function(gi) { %>
-            <info name="<%=gi["Property Name"]%>" value="<%=gi["Property Value"]%>"/>
+            <info name="<%=gi["Property Name"]%>" value="<%= _.escape(gi["Property Value"]) %> "/>
```
marcocast: Why we are escaping gi?

tobwiens: We put a json object inside the GI. the EXECUTION_CALENDARS object. That introduces thinks like < {} which will be interpreted as XML. That breaks the parsing of the workflow inside the Execution Scheduler

tobwiens: I tested Scheduler submission as well. It fails for the same reason:

Invalid workflow
At [tag:job] : Element type "info" must be followed by either attribute specifications, ">" or "/>".

MaelAudren: I think this an acceptable solution however we should be aware that if we want to validate the gi values as we do for variables, we will have to find another solution.

fviale:
The escape function must also be used in the Task Generic information definition.

PR #372, opened 2017-06-02
Enable to plan wf with variables

MaelAudren:


app/index.html
```diff
@@ -433,7 +433,7 @@
                     <div class="modal-footer">
                         <button class="btn btn-primary" id="check-button" type="button" aria-hidden="true">Check</button>
                         <button class="btn btn-primary" id="exec-button" type="button" aria-hidden="true">Execute</button>
-
+                        <button class="btn btn-primary" id="exec-plan-button" type="button" aria-hidden="true">Plan</button>
```
marcocast: no need to hide here

PR #378, opened 2017-06-21
make the workflow catalog portal show more than 20 buckets and workflows

sophiesongge:


sophiesongge:
yes @yinan-liu it is hard coded, it is just a work around, we have tried several methods to make pagination really work, but it didn't, and the client is waiting.

yinan-liu:
use PageRequest.java, we need two query params in the controller to receive the page number and size

sophiesongge:
thanks @yinan-liu  I'll try it again later

Issue #387, opened 2017-07-26
An error report is blocked in the popup of "workflow variables"


ruisong17:
When we try to define a workflow variable before creating any task in the ProActive Workflow studio, if we type in the Data type and then click on OK, there will be no response. We can see from console that an error is produced as shown in the snapshot below, but the error report in the popup is blocked.
<img width="557" alt="2017-07-25 2 38 56" src="https://user-images.githubusercontent.com/30046296/28620761-a865ab78-720e-11e7-852e-c766849b0f29.png">

PR #388, opened 2017-07-26
fix issue #387

ruisong17:
Removed a line that doesn't make sense in file studio-client.js. Now the error alert in the popup is no longer blocked.


app/scripts/proactive/model/Job.js
```diff
@@ -66,7 +66,7 @@ define(
                         if (StudioApp.isWorkflowOpen()) {
                             that.updateVariable(formValues);
                             var validationData = StudioClient.validate(StudioApp.views.xmlView.generateXml(), StudioApp.models.jobModel);
-                            if (!validationData.valid) {
+                            if (!validationData || !validationData.valid) {
```
fviale: I don't think this works. !validationData will return true if the validationData object is undefined. And if it's undefined, you will enter this block and will produce an error

PR #392, opened 2017-08-02
Fix the revision raw url in catalog view

paraita:
This PR should be merged after [this one](https://github.com/ow2-proactive/catalog/pull/44) as they are related.
This fixes the way the URL to a resource from the catalog view is generated.
As a proxy or a router can be setup in between the catalog and the client, we cannot rely on the URL being generated by the catalog service. Instead the browser will be able to get the relative path to the wanted resource on the catalog and it will be the responsability of the browser to generate the full URL which he should be able to since he's _talking_ to the proxy/router himself.


app/scripts/proactive/view/CatalogGetView.js
```diff
@@ -95,14 +95,18 @@ define(
...
-	        	var rawurl = $(currentRevisionRow).data("rawurl");
+	        	var rawurl = window.location.origin + '/catalog/' + $(currentRevisionRow).data("rawurl") + '/raw';
 	        	var name = $(currentRevisionRow).data("name");
 	        	var commitmessage = $(currentRevisionRow).data("commitmessage");
         		var projectName = $(currentRevisionRow).data("projectname");
         		
 	            this.highlightSelectedRow('#catalog-get-revisions-table', currentRevisionRow);
         		
 				var RevisionDescription = _.template(catalogRevisionDescription);
+				console.log('currentRevisionRow');
```
laurianed: Did you forget to remove your debug lines?

paraita: yes, thanks for noticing :)

PR #407, opened 2017-08-11
when publish new wf to catalog, refresh - get wf from catalog as current wf

ShatalovYaroslav:
So the generic info with groups will be added automatically to current workflow in the studio.


ShatalovYaroslav:
retest this please

PR #416, opened 2017-09-02
Persist workflow details automatically

zeineb:


zeineb:
Merge: Persist workflow details and generic information on change/update

Issue #410, opened 2017-08-18
Task Variables Are Ignored


tobwiens:
Having a simple task:

With code ``` echo " The task variable is: $variables_task_variable" ``` 

And following xml: 
```
<task name="Read_Task_Variables">
      <description>
        <![CDATA[ The simplest task, ran by a bash engine. ]]>
      </description>
      <variables>
        <variable name="task_variable" value="some_secret_stuff" inherited="false" model=""/>
      </variables>
      <scriptExecutable>
        <script>
          <code language="bash">
            <![CDATA[
env
echo " The task variable is: $variables_task_variable"
]]>
          </code>
        </script>
      </scriptExecutable>
    </task>
```

Works fine with the following output:
"The task variable is: some_secret_stuff "

But as soon as the workflow is exported and imported, the task XML changes to the following: 

```
<task name="Read_Task_Variables">
      <description>
        <![CDATA[ The simplest task, ran by a bash engine. ]]>
      </description>
      <variables>
        <variable name="task_variable" value="some_secret_stuff" inherited="true"  />
      </variables>
      <scriptExecutable>
        <script>
          <code language="bash">
            <![CDATA[
env
echo " The task variable is: $variables_task_variable"
]]>
          </code>
        </script>
      </scriptExecutable>
    </task>
```

One can see that ``` inherited="true"  ``` .....

The output is: "The task variable is: "

1) The inherited value should not change
2) The job has no such variable, why is the task variable deleted? 

Probably connected to: 
https://github.com/ow2-proactive/studio/issues/404
https://github.com/ow2-proactive/studio/issues/357

PR #420, opened 2017-10-02
 Fix bug of Inherited task variable checkbox is randomly turning on

### LINKED ISSUE #357: Inherited task variable checkbox is randomly turning on
In a middle of a workflow edition (it is very difficult to understand exactly when this happen), the inherited check box of a task variable is randomly auto-enabled

This basically breaks workflows from functioning properly as task variables are all-of-the-sudden not used.

luispinedamx:
This change resolves #357 and partially addresses #410 (which will come next).
Additionally fixes the random appearance of "model" attribute in the task variable tag of the XML, even when the value of "model" is empty.


paraita:
Can you squash or put more meaningful comments on your commits ? other than that LGTM :)

PR #421, opened 2017-10-05
Bug fixed : In window Get a Workflow from the Catalog: the list of buckets doesn't refresh automatically

chloegugli:


app/scripts/proactive/view/CatalogGetView.js
```diff
@@ -20,7 +20,7 @@ define(
             this.$el = $("<div id='catalog-get-container'></div>");
             $("#catalog-get-body").append(this.$el);
             this.buckets = options.buckets;
-            this.render();
+	    this.buckets.on('change', this.render, this);
```
marcocast: can you indent this line by one space?

PR #422, opened 2017-10-10
STUDIO BUG::In window Get a Workflow from the Catalog: the list of buckets doesn't refresh automatically

chloegugli:


paraita:
The commits comments are the same, can you put more details in those comments ? or simply squash the comments if the comments are note relevant please

paraita:
Switching this `async` to false isn't going to degrade the studio experience if the network is slow ?

chloegugli:
@paraita Yes, it might be noticeable by the user if the network is really slow

marcocast:
LGTM

PR #428, opened 2017-10-13
Selection scripts look like a list of clickable entries

chloegugli:


chloegugli:
@tobwiens It's a good idea but I would have to keep the check. In case type is undefined it will display 'selection script ()'

PR #433, opened 2017-10-24
add templates for native python task and change the name of jython ta…

sophiesongge:
…sk to jython


app/templates/script_cpython.xml
```diff
@@ -0,0 +1,32 @@
...
+    <task name="Python_Task">
+      <description>
+        <![CDATA[ The simplest task, ran by a python engine. ]]>
+      </description>
+      <scriptExecutable>
+        <script>
+          <code language="cpython">
+            <![CDATA[
+#By default the version of python you are using is the default python version on your machine
+#To change it to another version, you need to put a key 'PYTHON_COMMAND' in Generic Info,
```
tobwiens: I don't understand. I put the version of python into the PYTHON_COMMAND key? 

Can you then rename the PYTHON_COMMAND key to python version. How does it find the corresponding command on the linux machine? And how does it work when the machines are all setup differently?

sophiesongge: This is asked by Brian. The script engine executes the script by a command line like: python xxx.py, by default this 'python' is the python by default on your machine. If you have many versions of python, for example, you have at the same time python3, and you want to execute your script with 'python3 xxx.py', you just need to put 'python3' under the key PYTHON_COMMAND in GI, if python3 is not installed on your machine which means that 'python3 xxx,py' is not a correct command, then you will get an error as output saying that "python3 is not installed ...." and an error in the log with the same message. So it doesn't matter how the machines are set up, it's up to you, you know which python version you can use. The implementation part of this feature is in the project of jsr223-cpython.

tobwiens:
@sophiesongge Thank you for discussing the phrasing of the comment thoroughly with me. It is much more clear now. :+1:

PR #436, opened 2017-10-31
Load Templates menu from a default bucket

chloegugli:


app/scripts/proactive/view/WorkflowView.js
```diff
@@ -6,7 +6,7 @@ define(
         'proactive/view/TaskView',
         'proactive/view/utils/undo',
         'pnotify',
-        'pnotify.buttons'
+        'pnotify.buttons',
```
sophiesongge: Do we need a , here, since it is the last one?

chloegugli: No, you're right, it's not needed. Thank you

PR #441, opened 2017-11-03
New entry in "Catalog" menu to Set Bucket as Template Menu

chloegugli:


app/scripts/proactive/app.js
```diff
@@ -17,6 +17,7 @@ define(
         'proactive/view/LogoutView',
         'proactive/view/CatalogGetView',
         'proactive/view/CatalogPublishView',
+        'proactive/view/CatalogSetTemplatesBucketView',
         'proactive/view/WorkflowListView',
         'xml2json',
         'proactive/router',
```
sophiesongge: no need to have the , here

PR #448, opened 2017-11-10
Tasks with longer names improved

chloegugli:
Input for task name and task boxes with minimum and maximum sizes

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -186,7 +186,7 @@
         <table id="body-container" style="min-width:100%;height:100%">
             <tbody style="height:100%">
             <tr style="height:100%">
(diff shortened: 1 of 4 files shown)


app/index.html
```diff
@@ -186,7 +186,7 @@
         <table id="body-container" style="min-width:100%;height:100%">
             <tbody style="height:100%">
             <tr style="height:100%">
-                <td style="height:100%;min-width:300px;width:25%; padding-left:10px; padding-right:10px; padding-top:10px">
```
marcocast: is it possible to put this style conf in the css?

PR #451, opened 2017-11-13
Confirmation when removing properties (workflow variables, GI...)

chloegugli:
Message asking user if he really wants to remove the property

### DIFF
--- a/app/scripts/proactive/model/Job.js
+++ b/app/scripts/proactive/model/Job.js
@@ -82,6 +82,7 @@ define(
             "data-tab-help": "Workflow variables that will be available in all tasks.",
             "data-help": "<li><b>Name</b>: Name of the variable</li><li><b>Value</b>: Value of the variable</li>"
           },
+          confirmDelete: 'You are about to remove a variable.',
           itemToString: Utils.inlineNameValue,
           itemTemplate: bigCrossTemplate,
           subSchema: {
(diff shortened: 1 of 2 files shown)


app/scripts/proactive/model/Job.js
```diff
@@ -120,6 +121,7 @@ define(
           },
           itemToString: Utils.inlineNameValue,
           itemTemplate: bigCrossTemplate,
+          confirmDelete: 'You are about to remove a property.',
```
marcocast: Should not be a question? something like "are you really really sure you want to remove this property? :("

app/scripts/proactive/model/Task.js
```diff
@@ -105,7 +105,9 @@ define(
                             fieldAttrs: {'placeholder': '@attributes->accessMode'},
                             options: ["transferFromUserSpace", "transferFromGlobalSpace", "transferFromInputSpace", "transferFromOutputSpace","cacheFromUserSpace", "cacheFromGlobalSpace", "cacheFromInputSpace", "cacheFromOutputSpace", "none"]
                         }
-                    }
+                    },
+                    itemTemplate: bigCrossTemplate,
+                    confirmDelete: 'You are about to remove an input file.'
```
marcocast: same as before

app/scripts/proactive/model/Task.js
```diff
@@ -122,7 +124,9 @@ define(
                             fieldAttrs: {'placeholder': '@attributes->accessMode'},
                             options: ["transferToUserSpace", "transferToGlobalSpace", "transferToOutputSpace", "none"]
                         }
-                    }
+                    },
+                    itemTemplate: bigCrossTemplate,
+                    confirmDelete: 'You are about to remove an output file.'
```
marcocast: same as before

PR #454, opened 2017-11-22
Removable extra template menu

chloegugli:
Removing local templates (url /studio/#templates not available anymore)

Removable extra templates menu

New item in Catalog menu to open extra templates menus

Pointer cursor for catalog menu tables

Change palettes position to absolute when dragged (-> no more white space in menu)

### DIFF
--- a/app/scripts/proactive/view/dom.js
+++ b/app/scripts/proactive/view/dom.js
@@ -149,14 +149,27 @@ define(
             }
         });
 
-        $("#set-templates-bucket-button").click(function (event) {
+        $("#set-templates-main-bucket-button").click(function (event) {
             event.preventDefault();
             var studioApp = require('StudioApp');
             if (studioApp.isWorkflowOpen()){
                 studioApp.models.catalogBuckets.fetch({reset: true, async: false});
                 studioApp.modelsToRemove = [];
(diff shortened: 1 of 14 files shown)


app/styles/studio.css
```diff
@@ -85,6 +85,10 @@ footer {
     width: 80px;
 }
 
+.btn .caret {
+     margin-left: 4px
```
marcocast: the semicolon at the end...

app/scripts/proactive/view/dom.js
```diff
@@ -149,14 +149,27 @@ define(
...
-                studioApp.views.catalogSetTemplatesBucketView.render();
-                $('#set-templates-bucket-modal').modal();
+                studioApp.views.catalogSetMainTemplatesBucketView.render();
+                $('#set-templates-main-bucket-modal').modal();
+            }else{
+                $('#open-a-workflow-modal').modal();
+            }
+        });
+
+        $("#set-templates-secondary-bucket-button").click(function (event) {
```
marcocast: this function is the same as the previous one except that one is main and this is secondary. Do you think we can have a single external generic function that accepts a parameter to do one or the other?

PR #459, opened 2017-11-28
add execution for Url references in task implementation pre post clean script selection script and control flow

sophiesongge:


app/scripts/proactive/model/ControlFlow.js
```diff
@@ -0,0 +1,53 @@
...
+        schema: {
+            "Script": {
+                type: "TextArea",
+                fieldAttrs: {
+                    'placeholder': ['code->#cdata-section', 'code->#text']},
+                    template: scriptTemplate
+            },
+            "Language": {
+                type: 'Select',
+                options: [" ", "bash", "cmd", "docker-compose", "groovy", "javascript", "python", "cpython", "ruby", "perl", "powershell", "R"],
```
marcocast: one question, why cpython is called cpython and jython is called python?

sophiesongge: because jython is supplied by jsr-223, and it is called python, we can not change it

PR #464, opened 2017-12-01
replace icon by task.icon and fix issues

cdiop:


app/scripts/proactive/view/TaskView.js
```diff
@@ -172,9 +189,10 @@ define(
         	var iconPathGiValue;
             var genericInformation = this.model.get("Generic Info");
             for (var i in genericInformation) {
-                if (genericInformation[i]["Property Name"].toLowerCase() === 'icon') 
+                if (genericInformation[i]["Property Name"].toLowerCase() === 'task.icon') 
                 	iconPathGiValue = genericInformation[i]["Property Value"]; 
             }
+            this.model.get('Generic Info')['ICON.TASK'] = iconPathGiValue;
```
chloegugli: This 3 lines could go on the "if". Because if you add a GI that is not an icon, you'd remove the current task icon

PR #469, opened 2017-12-04
Lists that contain strings have the same remove button as other lists

chloegugli:
On the properties panel, lists that contain strings (Jvm Arguments, Additional Classpath & Arguments for native task implementation)  have the same remove button as other lists 
& only workflows in templates menu

### DIFF
--- a/app/scripts/proactive/model/utils.js
+++ b/app/scripts/proactive/model/utils.js
@@ -1,32 +1,38 @@
-define(function (require) {
+define(
+    [
+        'backbone'
+    ],
+
+    function (Backbone) {
 	
-    return {
-    	inlineName: function(prop) {
-            var name = prop['Name'] ? prop['Name'] : prop['Property Name'];
-            return name;
-        },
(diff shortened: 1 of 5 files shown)


app/scripts/proactive/model/utils.js
```diff
@@ -1,32 +1,38 @@
...
+                var checked = "";
+                if (prop['Inherited']){
+                    checked=" checked"
+                }
+
+                return "<input class='input-property-field' type=\"text\" value=\"" + prop['Name'] +
+                "\"><input class='input-property-field' type=\"text\"  value=\""+ value +
+                "\"><input class='input-property-field' type=\"checkbox\"" + checked + " onclick=\"return false;\">";
+            },
+            bigCrossTemplate : _.template('<div><span data-editor></span><button type="button" class="btn btn-danger" data-action="remove">x</button></div>')
```
fviale: probably not important but could you write it rather with bigCrossTemplate: (no space between the function name and the colon)

cdiop:
Oki For me

PR #479, opened 2017-12-15
display proper error message when importing not well-formed workflow

Aminelouati:


app/scripts/proactive/rest/studio-client.js
```diff
@@ -265,8 +265,12 @@ define(
         validateWithPopup: function (jobXml, jobModel, automaticValidation) {
             if (!localStorage['pa.session']) return;
             
-            if (automaticValidation && (jobModel.getTasksCount() == 0)) return;
+            if ( automaticValidation) {
```
marcocast: why this change?

Aminelouati: In javascript the evaluation of a boolean expression is atomic. When we import a not well-formed workflow, the job model is null in this case and therefore, jobModel.getTasksCount()  throws an exception. So this is a workaround to avoid this from happening.

PR #481, opened 2017-12-18
Enhance error message display when importing empty, not xml or not well-formed workflow

Aminelouati:


app/scripts/proactive/view/dom.js
```diff
@@ -97,15 +97,18 @@ define(
...
                     return;
                 }
                 var reader = new FileReader();
                 reader.onloadend = function (evt) {
 
                     if (evt.target.readyState == FileReader.DONE) {
                         var json = xml2json.xmlToJson(xml2json.parseXml(evt.target.result));
-                        StudioClient.validateWithPopup(evt.target.result, json, false);
+                        StudioClient.resetLastValidationResult();
+                        if(!StudioClient.validateWithPopup(evt.target.result, json, false)){
```
chloegugli: Why not putting the 3 lines (112 to 114) in the if and remove the '!' ?

PR #484, opened 2017-12-20
Keep the read from cache if username cookie variable is not defined

zeineb:


app/scripts/proactive/view/LoginView.js
```diff
@@ -22,10 +22,7 @@ define(
 
             fill: function() {
                 var username = this.getCookie('username');
-                if (username == "null") {
-                    $("#user").val("");
-                } else {
-
+                if (username != "null") {
```
marcocast: shoudn't be null without double quotes?

zeineb: If the cookie variable is not defined, it has this value "null", not null, or ""
So the check has to be done on this exact value "null"

zeineb: @marcocast this was the same remark by @Aminelouati :)

PR #486, opened 2017-12-22
Add PHP task

tobwiens:
- Add PHP task to 'Tasks' menu.
- Include icons


app/templates/php_script.xml
```diff
@@ -0,0 +1,73 @@
...
+        <task name="PHP_TASK">
+            <description>
+                <![CDATA[ Executes a php script. The PHP script be changed through the variable 'pathname' in the task variables. ]]>
+            </description>
+            <variables>
+                <variable name="pathname" value="script.php" inherited="false" />
+                <variable name="phpArguments" value="" inherited="false" />
+            </variables>
+            <genericInformation>
+                <info name="TASK.ICON" value="https://upload.wikimedia.org/wikipedia/commons/thumb/2/27/PHP-logo.svg/100px-PHP-logo.svg.png"/>
```
pedrovelho: Why commit the image if we get the logo from wikipedia?

tobwiens: It was part of the requirements.

PR #487, opened 2017-12-27
Notify the user when the job to submit contains EXECUTION_CALENDARS

zeineb:


zeineb:
I did a "javascript beautify" on the whole file, and this makes it difficult to know what was committed. To understand further what was changed:
- added a method isExecutionCalendarDefined (https://github.com/zeineb/studio/blob/d88f4386c700573f8dcf97389e43bec0e91cfea2/app/scripts/proactive/rest/studio-client.js#L225)
- invoked this method in "submit" (https://github.com/zeineb/studio/blob/d88f4386c700573f8dcf97389e43bec0e91cfea2/app/scripts/proactive/rest/studio-client.js#L203)

Issue #488, opened 2017-12-28
Cannot remove nameless tasks


amouhoub:
To reproduce this bug:
1. Create a task of any type
2. Edit the task name and remove its name
3. Try to remove it using the keyboard shortcut key `Delete` or `Suppr` (or `fn+Backspace` on Mac)
![capture d ecran 2017-12-28 a 15 46 13](https://user-images.githubusercontent.com/1918419/34414006-1ac333c8-ebe7-11e7-8adf-ea927fa60671.png)

- You will not be able to delete the task (see the attached picture)

amouhoub:
Fixed in PR #490

PR #490, opened 2017-12-29
FIX issue #488 related to empty task names

amouhoub:
In this PR, we fix the issue described in #488  that prevents nameless tasks from being removed unless they get a name.
The applied solution consists in notifying the user as shown in the attached screenshot.
**Note:** We followed the same behaviour as when the user gives a duplicate name to the task.
 
capture d ecran 2017-12-29 a 11 26 01

(Of course, the issue does not affect the scheduler because workflows with nameless tasks cannot be submitted to the scheduler).


mbenguig:
Tested and validated.

PR #491, opened 2017-12-29
Add alert method to handle task name update

zeineb:


app/scripts/proactive/view/TaskView.js
```diff
@@ -72,15 +72,35 @@ define(
...
                     
             this.element = $('<div class="task"><a class="task-name"><img src="'
-            	+ iconPath + '" width="20px">&nbsp;<span class="name">'
+                + iconPath + '" width="20px">&nbsp;<span class="name">'
                 + this.model.get("Task Name") + '</span></a></div>');
 
             this.showBlockInTask();
         },
 
+        alert: function(caption, message, type) {
```
amouhoub: I think that it would be better if you reuse the one (the same) declared in StudioClient, but this might be for later

PR #504, opened 2018-02-20
All portal buttons displayed (even current portal)

chloegugli:


app/index.html
```diff
@@ -150,30 +150,38 @@
...
+                            <form class="navbar-form" style="padding-right: 12px">
+                                <a class="btn btn-small btn-default" href="/automation-dashboard" target="/automation-dashboard"><img src="images/automation_dashboard_30.png" class="right-padding" style="height:25px">Automation Dashboard</a>
+                            </form>
+                        </div>
+                    </li>
+
+                    <li>
+                        <div>
+                            <form class="navbar-form" style="padding-right: 12px">
+                                <a class="btn btn-small btn-default active" href="/studio"><img src="images/studio_30.png" class="right-padding" style="height:25px">Workflow Studio</a>
```
luispinedamx: maybe add the _target_ to be uniform with the rest <a> tags ?

app/index.html
```diff
@@ -150,30 +150,38 @@
...
+                            </form>
+                        </div>
+                    </li>
+
                     <li>
                     <div>
                     	<form class="navbar-form" style="padding-right: 12px">
-                    		<a class="btn btn-small btn-default" href="/scheduler" target="_blank"><img src="images/scheduler_30.png"class="right-padding" style="height:25px">Scheduling & Orchestration</a>
+                    		<a class="btn btn-small btn-default" href="/scheduler" target="/scheduler"><img src="images/scheduler_30.png"class="right-padding" style="height:25px">Scheduling & Orchestration</a>
```
luispinedamx: So we will no longer open a new window?

app/index.html
chloegugli: It's the current portal, it needs to open in the current tab. So there is no need to add "target" attribute

app/index.html
chloegugli: This makes the link open in a new tab, and if the link has already been opened without closing it, it finds the right tab and re-opens it.

PR #517, opened 2018-03-07
Improve submission form

cdiop:


app/scripts/proactive/templates/job-variable-template.html
```diff
@@ -1,3 +1,29 @@
+<center>
+<% _.each(jobGenericInfos, function(gi) {
```
chloegugli: Is there a way to do that for each only one time to get all GI you need? Or why don't you use the same mechanism you used for jobProjectName, jobDescription .... ?

PR #531, opened 2018-04-12
alert about nameless workflows

Aminelouati:


MaelAudren:
Is it the expected behavior that the workflow can still be submitted even if it has no name ?

fviale:
I think it would be better to modify the server side to make a workflow invalid if it has an empty name

amouhoub:
@fviale @MaelAudren There are two sides to this anyway, and even if it is handled properly server-side (which must be done), it is good to display a nice warning message (Red popup, and red-glowing field for job-name) as a reminder to prevent the user from leaving it blank anyway

PR #537, opened 2018-04-20
fix encoding issue

Aminelouati:


app/scripts/proactive/view/dom.js
```diff
@@ -117,7 +117,7 @@ define(
 
                     }
                 }
-                reader.readAsBinaryString(file);
+                reader.readAsTextfile);
```
amouhoub: readAsTextfile();
Missing `(`

PR #539, opened 2018-05-07
Display task.documentation GI value as link

zeineb:


app/scripts/proactive/model/Task.js
```diff
@@ -381,11 +397,110 @@ define(
                 ]
 
                 this.controlFlow = {};
-                this.on("change", function (eventName, event) {
-                    undoManager.save()
+
+                // get the GI of the Task
+                var genericInformation = this.get["Generic Info"];
```
MaelAudren: I think it would be better to create a variable that will contain the value "Generic Info" instead of rewriting it each time.

app/scripts/proactive/model/Task.js
```diff
@@ -381,11 +397,110 @@ define(
...
+
+                        var genericInformation = eventName.changed["Generic Info"];
+                        var hasDocumentation;
+
+                        // Regenerate new documentation url
+                        for (var i in genericInformation) {
+                          if (genericInformation[i]["Property Name"].toLowerCase() === 'task.documentation') {
+                            hasDocumentation = true;
+                            var fileContent = "Documentation for the Job \"" + this.get('Name') + "\" \n" + "\n" + "\n";
+                            var fileContent = fileContent + "Documentation value: " + genericInformation[i]["Property Value"] + "\n";
```
MaelAudren: I think it would be better with a variable that contains the value "Property Value"

app/scripts/proactive/model/Task.js
```diff
@@ -381,11 +397,110 @@ define(
...
+
+                        if (hasDocumentation) {
+                          // Set New Value, needed for persistance
+                          this.set({
+                            "Generic Info Documentation": "{\"name\":\"" + linkName + "\",\"url\":\"" + documentationValue + "\"}"
+                          });
+
+                          // Needed to update UI
+                          $('a[name="Generic Info Documentation"]').text(documentationValue);
+                          $('a[name="Generic Info Documentation"]').attr("href", documentationValue);
```
MaelAudren: I think it would be better with to have a variable that concatenate a generic info variable and the string "Documentation"

app/scripts/proactive/model/Task.js
```diff
@@ -43,6 +43,16 @@ define(
                         "data-help": 'A small textual description of what task does.'
                     }
                 },
+                "Generic Info Documentation": {
+                  type: "Link",
+                  fieldAttrs: {
+                    "id": "gidoc",
+                    'placeholder': ['description->#cdata-section', 'description->#text'],
+                    "data-help": "Link to Task Documentation. Set value in Generic Information named \"DOCUMENTATION\""
```
Aminelouati: I think that the Generic Information is named \"task.documentation\"

PR #540, opened 2018-05-09
import workflow information (GI, Variables, Name, Project, Descriptio…n) when we drag&drop it from the palette

Aminelouati:
When we drag&drop a workflow (append) from the palette to the current workflow its information are not imported. In this PR, we add the elements that are not currently defined (No Name, or No Project or No Description, No GI or partial GI) to the current workflow.
So it also means to add the GI not yet defined.

### DIFF
--- a/app/scripts/proactive/model/Job.js
+++ b/app/scripts/proactive/model/Job.js
@@ -390,18 +390,23 @@ define(
         });
       },
       populate: function(obj, merging, isTemplate) {
-        // remove the unnecessary GI in case of appending workflow to the current one.
-        const GIToRemove = ["bucketname","documentation", "group", "pca.action.icon", "workflow.icon"];
-        if (isTemplate) {
-          if(obj["genericInformation"]){
-            if(obj["genericInformation"]["info"].isArray){
-              obj["genericInformation"]["info"] = obj["genericInformation"]["info"].filter(info => !GIToRemove.includes(info["@attributes"]["name"].toLowerCase()));
-            }
-            else{ // This is a workaround for "Controls tasks"
-              obj["genericInformation"]["info"] = []
-            }
-          }
-          this.populateTemplate(obj, merging);
+        var StudioApp = require('StudioApp');
+               // remove the unnecessary GI in case of appending workflow to the current one.
+               const COMMON_GI = ["bucketname","documentation", "group", "pca.action.icon", "workflow.icon"];
+               if (isTemplate) {
+                   if(obj["genericInformation"]){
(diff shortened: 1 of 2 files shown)


app/scripts/proactive/model/Job.js
```diff
@@ -390,18 +390,23 @@ define(
...
+               // remove the unnecessary GI in case of appending workflow to the current one.
+               const COMMON_GI = ["bucketname","documentation", "group", "pca.action.icon", "workflow.icon"];
+               if (isTemplate) {
+                   if(obj["genericInformation"]){
+                     var jobGenericInfos =  StudioApp.models.jobModel.get("Generic Info");
+                     var ExistingCommonGI = jobGenericInfos.filter(info => COMMON_GI.includes(info["Property Name"].toLowerCase())).map(a => a["Property Name"].toLowerCase());
+                     if(obj["genericInformation"]["info"] instanceof Array){
+                        obj["genericInformation"]["info"] = obj["genericInformation"]["info"].filter(info => !ExistingCommonGI.includes(info["@attributes"]["name"].toLowerCase()));
+                     }else{ // This is a workaround for "Controls tasks"
+                         var filtredGI = Object.keys(obj["genericInformation"]["info"]).map(name => obj["genericInformation"]["info"][name]).filter(info => !ExistingCommonGI.includes(info["name"].toLowerCase()));
```
zeineb: Very minor detail: You could have skipped the declaration of the variable filterGI, and test straightforwardly on "Object.keys(obj["genericInformation"]["info"]).map(name => obj["genericInformation"]["info"][name]).filter(info => !ExistingCommonGI.includes(info["name"].toLowerCase())).length"

app/scripts/proactive/model/Job.js
```diff
@@ -390,18 +390,23 @@ define(
...
-            }
-          }
-          this.populateTemplate(obj, merging);
+        var StudioApp = require('StudioApp');
+               // remove the unnecessary GI in case of appending workflow to the current one.
+               const COMMON_GI = ["bucketname","documentation", "group", "pca.action.icon", "workflow.icon"];
+               if (isTemplate) {
+                   if(obj["genericInformation"]){
+                     var jobGenericInfos =  StudioApp.models.jobModel.get("Generic Info");
+                     var ExistingCommonGI = jobGenericInfos.filter(info => COMMON_GI.includes(info["Property Name"].toLowerCase())).map(a => a["Property Name"].toLowerCase());
```
zeineb: You need to check if jobGenericInfos is empty or not.

Issue #505, opened 2018-02-21
Removing a Documentation generic info the link to user's guide does not restore


pedrovelho:
In a freshly created workflow the "General Parameters" shows a property Documentation with value "undefined" that points to users' guide section 3.2.

After creating a Generic Info -> "Documentation" with another value the link on the "General Parameters" tab changes. However after removing this Generic Info, the link goes back to undefined but does not point to the user guide anymore.

zeineb:
@pedrovelho I will fix this issue. I will let you know once it is done.

PR #555, opened 2018-06-04
Fix task.documentation link after GI removal #505

zeineb:
The commit fixes this issue https://github.com/ow2-proactive/studio/issues/505
It addresses also another bug: after deleting the task.documentation GI, clicking again on the task shows the deleted value.

### DIFF
--- a/app/scripts/proactive/model/Task.js
+++ b/app/scripts/proactive/model/Task.js
@@ -413,8 +413,6 @@ define(
                             var linkName = genericInformation[i]["Property Value"];
                             var documentationValue = genericInformation[i]["Property Value"];
                             break;
-                          } else {
-                            hasDocumentation = false;
                           }
                         }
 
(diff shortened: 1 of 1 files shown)


app/scripts/proactive/model/Task.js
```diff
@@ -459,17 +461,23 @@ define(
...
                   if (genericInformation[i]["Property Name"].toLowerCase() === 'task.documentation') {
+                    hasDocumentation = true;
                     linkName = genericInformation[i]["Property Value"];
                     documentationValue = genericInformation[i]["Property Value"];
                   }
                 }
-              } else {
-                console.log("No generic information task.documentation defined for this job");
+              } else{
+                console.log ("here");
```
luispinedamx: Maybe remove this logging?

zeineb: Thank you @luispinedamx, forgot to remove the debugging messages.

PR #570, opened 2018-07-09
Fix workflow variables being sorted alphabetically

### LINKED ISSUE #554: Variables are reordered alphabetically when saving a workflow
When creating a workflow, the automatic saving of the variables sorts them in alphabetic order, this is neither convenient (e.g. 1. UserName 2. Password; which in alphabetic order puts "Password" first), nor correct when we are trying to use variables in a specific order, e.g.
myVar : hello
luisVar : $myVar

this will sort luisVar first and myVar second, generating an error.

luispinedamx:
This PR corrects the bug of workflow variables being reordered alphabetically upon refreshing the studio or moving to another workflow.

fixes #554


luispinedamx:
@fviale exactly, the original array was being sorted.
@Aminelouati what I did notice is that this "findDuplicates" is only applied to workflow variables/generic info. Task variables/GI were not affected by this bug (i.e. the findDuplicates does not verifies them).

PR #571, opened 2018-07-13
Publish to catalog: be able to define the kind and the content type of the catalog object

zeineb:


app/scripts/proactive/view/dom.js
```diff
@@ -584,25 +584,27 @@ define(
...
             var payload = new FormData();
             payload.append('file', blob);
             payload.append('kind', 'workflow');
             payload.append('name', workflowName);
             payload.append('commitMessage', $("#catalog-publish-commit-message").val());
-            payload.append('objectContentType', "application/xml");
-            
+            payload.append('kind', $("#catalog-publish-kind").val());
+            payload.append('objectContentType', $("#catalog-publish-content-type").val());
+          //  payload.append('objectContentType', "application/xml");
```
chloegugli: Do you need this comment?

PR #573, opened 2018-07-25
Refactoring of script model management and presentation

fviale:
- moved all script classes to a dedicated folder. Script.js can now be configured on several aspects (help text, editor classes, etc).
 - refactor Script.js to use two nested models ScriptCode.js and ScriptFile.js. Radio button ScriptType allows to choose which model is active
 - dom.js: improved the TaskTypeRadioEditor handler to be able to display the corresponding nested model when the user click on one tab.
 - did a similar refactoring in Task.js, radio button Type selects the type of Task, and allows to activate one of the 3 available nested models ScriptExecutable, NativeExecutable, JavaExecutable
 - templates: made the corresponding changes due to the refactoring in existing templates
 - submodel-radio-form-template: new template related to the new radio button mechanism
 - BeautifiedModalAdapter: new class which allows some pre-treatment on the modal form displayed when editing selection scripts. This allows for example to display help tooltips.
 - backbone-forms: it was needed to do two small modifications inside the backbone forms code to write the submodel-radio-form-template and to display default ScriptType on selection scripts modal window. Backbone-forms is not actively developed any more and when trying to upgrade our version 0.14.0 to 0.14.1, the whole form was broken.


app/libs/backbone-forms/distribution.amd/backbone-forms.js
```diff
@@ -825,7 +825,8 @@ Form.Field = Backbone.View.extend({
       fieldAttrs: schema.fieldAttrs,
       editorAttrs: schema.editorAttrs,
       key: this.key,
-      editorId: this.editor.id
+      editorId: this.editor.id,
+      self: this
```
chloegugli: Are you sure it's a good idea to modify the library code? You couldn't do anything else?

app/scripts/proactive/view/BeautifiedModalAdapter.js
```diff
@@ -0,0 +1,86 @@
...
+        'backboneFormsAdapter',
+        'StudioApp',
+        'bootstrap',
+    ],
+
+    function ($, Backbone, PNotify) {
+
+        "use strict";
+
+        // Extension of the bootstrap modal form to operate some cosmetic operations on the form.
```
chloegugli: Did you make sure this doesn't removed the style we have already done for the form? (spacing between fields, button colors...)

app/scripts/proactive/view/TaskView.js
```diff
@@ -285,6 +277,8 @@ define(
 
             if (!model) return;
 
+            model.model.parentModel = this.model;
```
chloegugli: model.model.parentModel: there are a lot of models, is it normal to put a model in the model?

app/styles/studio.css
```diff
@@ -681,6 +682,42 @@ input[type="checkbox"] {
...
+    margin-top: 18px;
+    font-weight: bold !important;
+}
+
+/* buttons for arguments in selection script modal */
+div[name="Arguments"] button {
+  font-size: 12px;
+}
+
+.width100p {
```
chloegugli: Usually class name represent what the element does or the context

paraita: I agree !

app/libs/backbone-forms/distribution.amd/backbone-forms.js
paraita: it was a bad move from the begining to version the libraries but we can't revert without breaking stuff so I guess it's ok for now. We can think of a better solution in the future.

fviale: I resolved to this as a last resort after exploring all possible solutions. Even tried to upgrade backbone-forms. But after the upgrade the whole forms were broken. The code seems to do already a lot of custom stuff.
It's relatively safe in the sense that backbone-forms is not maintained any more, so we don't risk being completely out-of-date.

app/libs/backbone-forms/distribution.amd/editors/list.js
```diff
@@ -539,7 +549,7 @@ define(['jquery', 'underscore', 'backbone', 'backbone-forms'], function($, _, Ba
     onModalSubmitted: function() {
       var modal = this.modal,
           form = this.modalForm,
-          isNew = !this.value;
+          isNew = !this.value || (this.defaultValue && JSON.stringify(this.defaultValue) === JSON.stringify(this.value));
```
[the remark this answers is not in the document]
fviale: ok, I can rename this variable

app/scripts/proactive/model/ForkEnvironment.js
```diff
@@ -64,10 +63,13 @@ define(
                     type: 'NestedModel',
                     model: ForkEnvironmentScript,
                     fieldAttrs: {
-                        'placeholder': 'envScript->script',
-                        "data-help": 'Environment script that is able to add/change each items of the fork environment programmatically.'
+                        'placeholder': 'envScript',
+                        "data-help": 'A script which can be used to configure programmatically the Java Virtual Machine forked from the computing node to execute a task. The environment script is run before any other task script (pre, task, etc).'
```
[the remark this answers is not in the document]
fviale: ok

app/scripts/proactive/view/BeautifiedModalAdapter.js
fviale: This operation (beautifyForm) is already applied to the left panel form (in class ViewWithProperties). It adds all necessary classes to the fields, implement tooltips, etc. The problem was this operation was not performed on the modal window when we want to edit selection scripts. This new class allows it.

app/scripts/proactive/view/TaskView.js
fviale: I can work to make this code more understandable, yes.

PR #582, opened 2018-08-21
Changes after validation

chloegugli:
Validation didn't succeed because:
-It was not possible to import pca workflows
-Groovy scripts were displayed as text in code mirror of catalog portal
-Some catalog objects didn't have extensions


app/scripts/proactive/config.js
```diff
@@ -95,17 +95,17 @@ define(function () {
         },
         languages_content_type: {
             "scalaw" : "text/x-scala",
-            "groovy" : "application/x-groovy",
-            "javascript" : "application/javascript",
-            "python" : "application/x-python",
-            "cpython" : "application/x-python",
-            "ruby" : "application/x-ruby",
+            "groovy" : "text/x-groovy",
+            "javascript" : "text/javascript",
```
amouhoub: @chloegugli Why are they becoming text/xxxxx now? (I know it is a common debate on whether this is meant to be manipulated by machine or read by humans)

app/scripts/proactive/config.js
```diff
@@ -95,17 +95,17 @@ define(function () {
...
+            "python" : "text/x-python",
+            "cpython" : "text/x-python",
+            "ruby" : "text/x-ruby",
             "r" : "text/x-R",
             "bash" : "application/x-sh",
```
amouhoub: depending on the logic upon which you changed the other from `application` to `text` MIME types, these ones should also follow that logic

app/scripts/proactive/config.js
chloegugli: Actually this is a fix for "-Groovy scripts (and others) were displayed as text in code mirror of catalog portal". So I had a look at which content-types code mirror understands : https://codemirror.net/mode/

app/scripts/proactive/config.js
chloegugli: After looking at which content-types code mirror understands (https://codemirror.net/mode/), some are 'text/' (groovy for example) and some are 'application' (powershell for example)

amouhoub: 👍

app/scripts/proactive/config.js
amouhoub: 👍

PR #587, opened 2018-08-29
Fix duplicated revisions and commit message in import/export modals

chloegugli:


app/scripts/proactive/view/CatalogGetView.js
```diff
@@ -254,8 +255,12 @@ define(
         },
         render: function () {
             this.$el.html(this.template());
-            this.updateBuckets();
-            this.buckets.on('sync', this.updateBuckets, this);
+            var bucketKind = this.kind;
+            //for workflows, we don't want subkind filters (ie we want to be able to import workflow/pca and workflow/standard)
+            if (this.kind.toLowerCase().indexOf('workflow') > -1)
```
fviale: always better to use curly brackets for code clarity

app/scripts/proactive/view/CatalogPublishView.js
```diff
@@ -202,8 +203,12 @@ define(
         },
         render: function () {
             this.$el.html(this.template());
-            this.updateBuckets();
-            this.buckets.on('sync', this.updateBuckets, this);
+            var bucketKind = this.kind;
+            //for workflows, we don't want subkind filters
+            if (this.kind.toLowerCase().indexOf('workflow') > -1)
```
fviale: same comment as above

fviale: not sure this is robust as you search for workflow anywhere in the string.
If the workflow kind are:
workflow/standard and workflow/pca

To avoid conflicts with user-defined kinds, it would probably be better to use:
if (this.kind.toLowerCase().indexOf('workflow') == 0) {
}
We now the workflow kinds starts with workflow

app/scripts/proactive/view/CatalogGetView.js
```diff
@@ -56,8 +57,9 @@ define(
                 var bucketName = that.getSelectedBucketName();
                 var filterKind = this.kind;
                 //for workflows, we don't want subkind filters (ie we want to be able to import workflow/pca and workflow/standard)
-                if (this.kind.toLowerCase().indexOf('workflow') > -1)
-                    filterKind = "workflow"
+                if (this.kind.toLowerCase().indexOf('workflow') > -1) {
```
fviale: not sure this is robust as you search for workflow anywhere in the string.
If the workflow kind are:
workflow/standard and workflow/pca

To avoid conflicts with user-defined kinds, it would probably be better to use:
if (this.kind.toLowerCase().indexOf('workflow') == 0) {
}
We know the workflow kinds starts with workflow

chloegugli: You're right, I'll change that

PR #591, opened 2018-09-06
Be able to see and modify the code of imported reference

chloegugli:


chloegugli:
@tobwiens The thing is that we put in dom.js all the code to execute when the document is ready. If we take the example of a listener, the function executed inside it could be moved into a specific view. This would reduce the size of the file but all listeners would still be set in dom.js. Also, there might be code duplication to remove.

PR #593, opened 2018-09-10
Be able to retrieve imported/exported script more easily when exporting

chloegugli:


app/scripts/proactive/view/CatalogGetView.js
```diff
@@ -213,16 +213,19 @@ define(
...
                 $('#catalog-get-close-button').click();
                 StudioClient.alert('Import successful', 'The ' + that.kindLabel + ' has been successfully imported from the Catalog', 'success');
-                //if it's a script, we set the language depending on the file extension
                 if (that.kind.toLowerCase().indexOf('script') == 0) {
+                    //saving script name and bucket for next commits
+                    inputToImport.dataset.scriptName = that.getSelectedObjectName();
+                    inputToImport.dataset.bucketName = that.getSelectedBucketName();
                     try {
+                        //if it's a script, we set the language depending on the file extension
                         var contentDispositionHeader = request.getResponseHeader('content-disposition');
```
fviale: content-disposition is only available when downloading from the catalog. If it's not present, could you use the url directly instead ?

chloegugli: This function is only called when importing from catalog

fviale: Ok my mistake

PR #604, opened 2018-10-02
Create default php-script for PHP task in pre script.

tobwiens:
Intention:
- The PHP Task should succeed when executed as default

Description:
- Add PHP code in the pre-script
- User which drags the PHP task and executes it, will have the task working immediately with default settings.


tobwiens:
@Mykhailenko @ShatalovYaroslav Thank you very much

PR #606, opened 2018-10-08
Fix Shell template and reorder tasks

Mykhailenko:


sophiesongge:
Although, I don't know why the order of the tasks need to be reordered :)

PR #616, opened 2018-11-02
Display committed by username when importing from catalog

amouhoub:
In this PR, we display the username that committed the objects/wfs to catalog in the import object from catalog modals.

### DIFF
--- a/app/scripts/proactive/view/CatalogGetView.js
+++ b/app/scripts/proactive/view/CatalogGetView.js
@@ -151,7 +151,7 @@ define(
 	        	var name = $(currentRevisionRow).data("name");
 	        	var commitMessage = $(currentRevisionRow).data("commitmessage");
         		var projectName = $(currentRevisionRow).data("projectname");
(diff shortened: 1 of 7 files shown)


app/scripts/proactive/view/CatalogGetView.js
```diff
@@ -149,16 +149,18 @@ define(
             if (currentRevisionRow){
 	        	var rawurl = window.location.origin + '/catalog/' + $(currentRevisionRow).data("rawurl") + '/raw';
 	        	var name = $(currentRevisionRow).data("name");
-	        	var commitmessage = $(currentRevisionRow).data("commitmessage");
+	        	var commitMessage = $(currentRevisionRow).data("commitmessage");
         		var projectName = $(currentRevisionRow).data("projectname");
+        		var username = $(currentRevisionRow).data("username");
```
luispinedamx: the var could be `userName` for consistency with the previous vars

PR #621, opened 2018-11-23
Change to cope with changes on server.

pedrovelho:
Stop working because now the REST call returns success from the server side triggering the right callback on the javascript.

### DIFF
--- a/app/scripts/proactive/rest/studio-client.js
+++ b/app/scripts/proactive/rest/studio-client.js
@@ -80,7 +80,10 @@ define(
                     type: 'POST',
                     success: function(data) {
                         // ProActive Studio login request return invalid json with status code 200
-                        console.log("Should not be there", data)
+                        that.alert("Connected", "Successfully connected user", 'success');
+                        console.log("Session ID is " + data.responseText)
+                        localStorage['pa.session'] = data.responseText;
+                        return onSuccess();
                     },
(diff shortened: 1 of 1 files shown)


app/scripts/proactive/rest/studio-client.js
```diff
@@ -80,7 +80,10 @@ define(
                     type: 'POST',
                     success: function(data) {
                         // ProActive Studio login request return invalid json with status code 200
-                        console.log("Should not be there", data)
+                        that.alert("Connected", "Successfully connected user", 'success');
```
chloegugli: This alert might be annoying

app/scripts/proactive/rest/studio-client.js
```diff
@@ -80,7 +80,10 @@ define(
...
+                        console.log("Session ID is " + data.responseText)
+                        localStorage['pa.session'] = data.responseText;
```
chloegugli: This function says (in the comment that it sets localStorage['pa.login']. Why setting localStorage['pa.session'] here? Moreover, you use it in the request header.

amouhoub:
@pedrovelho what about this PR ?

pedrovelho:
I belive this was not to blame at the end for the problem! The problem was some guys that had a plugin installed in their browsers: does it ring a bell?

Issue #632, opened 2018-12-18
Issue when creating variables with incorrect names.


fviale:
Issue appear when creating a variable with an incorrect name and a Model is defined.

First step (which is ok), an error message appears:

![image](https://user-images.githubusercontent.com/3063713/50146546-4797a480-02b4-11e9-8d54-99c2336373a9.png)

When trying to fix the variable name, the error message does not disappear:

![image](https://user-images.githubusercontent.com/3063713/50146613-70b83500-02b4-11e9-9a21-a558711d07eb.png)

After cancelling the form and reloading, several variables appear in the list instead of one:
![image](https://user-images.githubusercontent.com/3063713/50146687-a0673d00-02b4-11e9-9cdd-3749186eda99.png)

amouhoub:
@fviale is this fixed by the aforementioned commits of yours ?

PR #633, opened 2018-12-18
Fix for #632 Issue when creating variables with incorrect names.

fviale:
- store original variables when we are validating the variable form (BackupVariables). Any modification will only apply to these original variables and will not pile up (which happens when we change the variable name inside the form at each unsuccessful attempt).
 - when the form is valid, remove this BackupVariables attribute.


app/scripts/proactive/model/Job.js
```diff
@@ -332,17 +334,23 @@ define(
         }
 
         if (this.attributes.hasOwnProperty('Variables')) {
-          var variables = this.attributes.Variables;
+          var variables;
+          if (this.attributes.hasOwnProperty('BackupVariables')) {
+            this.attributes.Variables = JSON.parse(JSON.stringify(this.attributes.BackupVariables));
+          } else {
+            this.attributes.BackupVariables = JSON.parse(JSON.stringify(this.attributes.Variables));
```
chloegugli: Can you add comments saying why you need this.attributes.BackupVariables and what you're doing with it?

app/scripts/proactive/model/Job.js
```diff
@@ -332,17 +334,23 @@ define(
...
+          }
+          variables = this.attributes.Variables;
           var index = -1
           for (var i = 0; i < variables.length; i++) {
             if (variables[i].Name == variable.Name) {
               index = i;
```
chloegugli: we could maybe add a break here

app/scripts/proactive/model/Task.js
```diff
@@ -562,17 +564,23 @@ define(
...
+                    if (this.attributes.hasOwnProperty('BackupVariables')) {
+                      this.attributes.Variables = JSON.parse(JSON.stringify(this.attributes.BackupVariables));
+                    } else {
+                      this.attributes.BackupVariables = JSON.parse(JSON.stringify(this.attributes.Variables));
+                    }
+                    variables = this.attributes.Variables;
                     var index = -1
                     for (var i = 0; i < variables.length; i++) {
                         if (variables[i].Name == variable.Name) {
                             index = i;
```
chloegugli: Same comment about the break

app/scripts/proactive/model/Task.js
```diff
@@ -562,17 +564,23 @@ define(
                     variable.Value = "";
                 }
                 if (this.attributes.hasOwnProperty('Variables')) {
-                    var variables = this.attributes.Variables;
+                    var variables;
+                    if (this.attributes.hasOwnProperty('BackupVariables')) {
+                      this.attributes.Variables = JSON.parse(JSON.stringify(this.attributes.BackupVariables));
```
chloegugli: Same comment about BackupVariables

app/scripts/proactive/model/Job.js
fviale: done

app/scripts/proactive/model/Job.js
fviale: done

app/scripts/proactive/model/Task.js
fviale: done

app/scripts/proactive/model/Task.js
fviale: done

PR #636, opened 2019-01-09
Use textarea in variable values

fviale:
When entering long values in variables, as the input element width is not resizable, it is complicated to view the value entered or spot mistakes.

This changes the input field used for variable-like values (Job and Task variables, Generic information, and job submission) from text to textarea, allowing to resize the field and thus better view the entered value.

Additionally:
 - make all textarea resizable vertically only (resizing horizontally creates issues on the studio).
 - change the label "Number of Automatic Restarts", which is incorrect, to "Number of Execution Attempts"


app/styles/studio.css
```diff
@@ -517,10 +517,16 @@ ANALOGIC SCHEME
     min-height: 2em !important;
 }
 
+.textareavalues {
+    min-height: 1em !important;
+    min-width: unset;
```
chloegugli: Why did you do ' min-width: unset;' ?

fviale: otherwise it inherits from min-width: 90% which does not align with other fields in the modal form

fviale: but maybe there is a better approach than unset

chloegugli: Another solution would be to look at the CSS of the other fields. But this seems OK too.

fviale: ok, I will leave it like this then.

PR #638, opened 2019-01-21
Improve Calalog palette with new features and bug fixes

amouhoub:
In this PR:
- Add pin functionality to pin/unpin the palette(sticky/relative position).
- Add button to add buckets to the palette easily
- Reconstruct the tools panel into a 3 div column design instead of cell design
- New buckets are added in a new line if width is insufficient
- fix old bugs like dragging an element under the left panel
- improve the layout for multiple things

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -202,20 +202,30 @@
                 <td id="left-panel-properties">
                     <div id="properties-container"></div>
                 </td>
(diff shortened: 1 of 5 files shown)


amouhoub:
thx @chloegugli @fviale .
It was also validated by other folks on trydev directly,,, as you said, nice way to review by testing !!!

PR #643, opened 2019-02-06
Add Presets feature

amouhoub:
In this PR, we add a new feature to the studio called the Presets.
It allows to load a predefined set of buckets in the Palette toolbar.

More precisely, this PR:
1) Refactors the old palette rendering fashion based on main+secondary bucket menues
2) Adds a preset definition in the config.js
3) Adds a preset loading menu from View=>Change Studio Preset
4) Adds a new modal for loading presets.
7) Beautifies many files (HTML, javascript)

### DIFF
--- a/app/scripts/proactive/view/dom.js
+++ b/app/scripts/proactive/view/dom.js
@@ -215,29 +215,31 @@ define(
             }
         });
 
-        function openSetTemplatesMenuModal(order){
+        function openAddPaletteBucketMenuModal(){
             var studioApp = require('StudioApp');
             if (studioApp.isWorkflowOpen()){
                 studioApp.models.catalogBuckets.setKind("workflow");
                 studioApp.models.catalogBuckets.fetch({reset: true, async: false});
-                studioApp.modelsToRemove = [];
-                if (order=='main')
-                    studioApp.views.catalogSetMainTemplatesBucketView.render();
-                else if (order=='secondary')
-                    studioApp.views.catalogSetSecondaryTemplatesBucketView.render();
-                $('#set-templates-'+order+'-bucket-modal').modal();
+                studioApp.views.catalogSetSecondaryTemplatesBucketView.render();
+                $('#set-templates-secondary-bucket-modal').modal();
             }else{
                 $('#open-a-workflow-modal').modal();
             }
         }
-        $("#set-templates-main-bucket-button").click(function (event) {
-            event.preventDefault();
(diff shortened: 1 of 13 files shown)


app/scripts/proactive/templates/catalog-set-templates-bucket.html
```diff
@@ -5,7 +5,7 @@
                 <fieldset>
                     <legend style="font-weight: bold;">Buckets</legend>
                     <div style="height: 270px; overflow: auto;">
-                        <table id="catalog-set-templates-<%- this.order %>-bucket-table" style="width: 100%; cursor: pointer;"></table>
+                        <table id="catalog-set-templates-secondary-bucket-table" style="width: 100%; cursor: pointer;"></table>
```
chloegugli: Are you sure this won't result in several elements with the same ID?

app/scripts/proactive/view/dom.js
```diff
@@ -288,7 +290,26 @@ define(
...
             var bucketName = ($(($("#catalog-set-templates-secondary-bucket-table .catalog-selected-row"))[0])).text();
-            require('StudioApp').views.palleteView.setSecondaryTemplatesBucket(bucketName, false);
+            // Leave modal open if bucket could not be added (normally because it's already in the Palette)
+            if (require('StudioApp').views.paletteView.addPaletteBucketMenu(bucketName, false)){
+                $('#set-templates-secondary-bucket-modal').modal('hide');
+            }
+        });
+
+        // Set default preset on startup
+        if (!localStorage['palettePreset']){
```
chloegugli: I don't see where you use localStorage['palettePreset']. What happens if buckets that are saved in localStorage['palettePreset'] (or config.palette_presets) are removed?

app/scripts/proactive/templates/catalog-set-templates-bucket.html
amouhoub: no, because this is a table, and there is only one table in the Add bucket to palette modal
Besides, `<%- this.order %>` used to be either `main` or `secondary`.
There is no `main buckets in the palette anymore` you can freely do whatever you want.
the only menu you can't remove is `Tasks`

app/scripts/proactive/view/dom.js
amouhoub: they are used in `PalettePreview.js`

amouhoub:
retest this please

PR #650, opened 2019-02-25
Do copy/past of task between workflows.

medou-boushab:


app/scripts/proactive/view/WorkflowView.js
```diff
@@ -509,7 +509,7 @@ define(
                 return (name === view.model.get("Task Name"));
             })
         },
-        copyPasteTasks: function (tasks, position) {
+        copyPasteTasks: function (tasks, position , newTaskModels, Taks) {
```
MaelAudren: I don't understand what is `Taks`, maybe you could rename it into something more explicit. I think that you should follow the naming convention and put the variable in lowercase.

app/scripts/proactive/view/dom.js
```diff
@@ -1027,16 +1027,24 @@ define(
...
                         copied.push(t);
+                        task.push($(t).data( "view" ))
                     })
                 }
                 if (ctrlDown && e.keyCode == vKey) {
                     if (pasteAllow) {
-                        console.log("paste");
-                        require('StudioApp').views.workflowView.copyPasteTasks(copied, pasteAllow);
+                        var newTaskModel = []
+                        var Tasksview = [];
```
MaelAudren: Please follow the naming convention and rename `Tasksview` in `tasksView`

app/scripts/proactive/view/dom.js
```diff
@@ -1027,16 +1027,24 @@ define(
...
+                         $.each(copied, function (i, t) {
```
chloegugli: As you loop in copied to get elements for task, maybe you don't need copied table anymore. Also, it seems that there are several tasks in 'task' so you might rename it 'tasks'

app/scripts/proactive/view/WorkflowView.js
fviale: looks like a typo, but yes it should be more understandable. for example tasksToCopy, tasksFromCurrentModel (I don't know the precise semantic here)

app/scripts/proactive/view/dom.js
```diff
@@ -998,7 +998,7 @@ define(
         })();
 
         $(document).ready(function () {
-
+            var copiedTask = [];
```
chloegugli: There are several tasks inside copiedTask right? 'copiedTasks' might be more appropriate

app/scripts/proactive/view/WorkflowView.js
medou-boushab: done

PR #657, opened 2019-03-04
Workflow Submit: Automatically check when changing a value in list or radio button

fviale:


app/scripts/proactive/templates/job-variable-template.html
```diff
@@ -45,7 +45,7 @@ <h4 class="modal-title"><%- jobName %> </h4>
                     <% } %>
                 </select>
             <% } else if (variable.hasOwnProperty("Model") && variable.Model.toLowerCase().indexOf('pa:boolean')!=-1) {  %>
-                <div class="variableValue radio-div" id="<%- key %>" data-variable-model="<%- variable.Model %>">
+                <div class="variableValue radio-div" id="<%- key %>" onchange="$('#check-button').click()" data-variable-model="<%- variable.Model %>">
```
chloegugli: Are you sure about this onchange attribute on a DIV ?

fviale: I tested it and it seems to work.

PR #660, opened 2019-03-06
Select all tasks with Ctr-A

medou-boushab:


app/scripts/proactive/view/dom.js
```diff
@@ -1057,6 +1057,10 @@ define(
                 if (ctrlDown && e.keyCode == yKey) {
                     undoManager.redoIfEnabled();
                 }
+                if(ctrlDown  && e.keyCode == aKey ){
+                       e.preventDefault();
+                       $(".task").addClass("selected-task");
```
chloegugli: Don't you need to update the list of selected tasks too? If you want to do copy paste for example

medou-boushab: No, we don't need to update it. We use select them by $(".selected-task").

PR #661, opened 2019-03-07
Select all

medou-boushab:


app/scripts/proactive/view/dom.js
```diff
@@ -1016,17 +1016,16 @@ define(
 
 
             var ctrlDown = false;
-            var ctrlKey = 17, commandKey = 91, vKey = 86, cKey = 67, zKey = 90, yKey = 89;
+            var ctrlKey = 17, commandKey = 91, vKey = 86, cKey = 67, zKey = 90, yKey = 89, aKey = 65;
             var pasteAllow = true;
             var canDoPast = false
-
-            $(document).keydown(function (e) {
+            $('#workflow-designer-outer').bind('keydown', function (e) {
```
fviale: I am afraid by doing this change you disable copy/paste at all outside of the canvas

medou-boushab: No, it still works

fviale: you are right, just testing it right now and it seems to work as expected. Before approve it, I will check with you tomorrow so I can understand the changes.

app/scripts/proactive/view/dom.js
```diff
@@ -1019,14 +1019,13 @@ define(
...
                 if (e.keyCode == ctrlKey || e.keyCode == commandKey) ctrlDown = true;
             }).keyup(function (e) {
                 if (e.keyCode == ctrlKey || e.keyCode == commandKey) ctrlDown = false;
             });
 
-            $(document).keydown(function (e) {
+            $('#workflow-designer-outer').bind('keydown',function (e) {
```
fviale: could you change the title of the pop-up ?
instead of "alert-success" => "Copy / Paste"

medou-boushab: done

PR #663, opened 2019-03-12
Allow the workflow to go back to its previous state by Ctrl-Z

medou-boushab:


app/scripts/proactive/view/utils/undo.js
```diff
@@ -37,17 +37,21 @@ define(
...
+        undo: function (num) {
             if (undoStates.length <= 1) {
                 StudioClient.alert("No further undo data", "");
                 return;
             }
-            this._move(undoStates, redoStates);
-            this._restoreLastState();
+            var length = undoStates.length;
+            for(var i= 1; i<num + 1; i++){
+            undoStates = undoStates.slice(0,length - i + 1);
```
chloegugli: Can you indent this please?

app/scripts/proactive/view/utils/undo.js
```diff
@@ -37,17 +37,21 @@ define(
...
+                this._move(undoStates, redoStates);
+                this._restoreLastState();
+            }
         },
-        undoIfEnabled: function () {
+        undoIfEnabled: function (numTasks) {
             if (enabled) {
-                this.undo();
+                this.undo(numTasks);
```
chloegugli: It seems that you narrowed down the use of this function for undoing tasks creation. It can be used for other things, like removing a newly created variable for example.

app/scripts/proactive/view/utils/undo.js
```diff
@@ -37,17 +37,28 @@ define(
...
         },
-        undo: function () {
+        undo: function (num) {
             if (undoStates.length <= 1) {
                 StudioClient.alert("No further undo data", "");
                 return;
             }
-            this._move(undoStates, redoStates);
-            this._restoreLastState();
+            // We want to know where we call this function: undo of workflow variables or tasks
```
chloegugli: This comment is misleading, there are many different undo, not only workflow variables or tasks.

medou-boushab: that' right, I'll update it :)

PR #681, opened 2019-06-18
Bug: when trying to drag & drop a task 

medou-boushab:


ShatalovYaroslav:
In this case as it brakes the import from catalog, it's better to leave as it is now and don't merge.

PR #687, opened 2019-07-12
Studio:: Add Copy, Paste, Cut, select All and delete entries in menu Edit

medou-boushab:
1- Able to do  Ctrl+X
2- Create a menu: Add Copy, Paste, Cut, select All and delete,

### DIFF
--- a/app/scripts/proactive/view/ViewWithProperties.js
+++ b/app/scripts/proactive/view/ViewWithProperties.js
@@ -62,7 +62,8 @@ define(
 					} else {
 						jobBreadcrumb = $('<li class="active"><span id="breadcrumb-job-name">' + modelInUse.get("Name") + '</span></li>');
 						breadcrumb.append(jobBreadcrumb)
-
+                        //hidden the context-menu
+                        $(".context-menu-canvas").hide();
 						// selected-task class is used for copy/paste, delete operations, group task moving etc
 						$(".selected-task").removeClass("selected-task");
 						// active-task class is used to identify which task is currently shown in the properties view
--- a/app/scripts/proactive/view/ViewWithProperties.js
+++ b/app/scripts/proactive/view/ViewWithProperties.js
@@ -314,7 +314,7 @@ define(
 						var that = this;
 						var removeTask = $('<a href="#" class="btn btn-danger btn-sm pull-right bottom-btn" title="Remove task">Remove task</a>');
 						removeTask.click(function () {
-							StudioApp.views.workflowView.removeView(that);
+							StudioApp.views.workflowView.removeViewWithDependencies(that);
 							return false;
 						})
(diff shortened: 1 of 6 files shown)


app/scripts/proactive/view/ViewWithProperties.js
```diff
@@ -62,7 +62,8 @@ define(
 					} else {
 						jobBreadcrumb = $('<li class="active"><span id="breadcrumb-job-name">' + modelInUse.get("Name") + '</span></li>');
 						breadcrumb.append(jobBreadcrumb)
-
+                        //hidden the context-menu
+                        $(".context-menu-canvas").hide();
```
MaelAudren: Why the indentation is not at the same level ?

app/scripts/proactive/view/dom.js
```diff
@@ -689,18 +689,26 @@ define(
             var studioApp = require('StudioApp');
             studioApp.views.catalogPublishView.publishToCatalog();
         })
-
+        // The aim object of this function is to delete sected tasks
```
MaelAudren: sected => selected

app/scripts/proactive/view/ViewWithProperties.js
medou-boushab: I don't know why.

MaelAudren: Can you fix it ?

app/scripts/proactive/view/dom.js
```diff
@@ -1037,43 +1045,90 @@ define(
...
+                })
+                // let the user how he can do past(ctr-v)
+                if(copiedTasks.length > 0){
+                    StudioClient.alert('Copy/Paste', 'Click on the canvas where you want to paste and then do ctrl-V.', 'warning');
+                } else {
+                    StudioClient.alert('Copy/Paste', 'Select at least one task.', 'warning');
+                }
+
+                console.log('positions')
+                console.log(positions)
```
MaelAudren: Not sure that those log are required and if you thinks that they are required should put them in one line.

app/scripts/proactive/view/ViewWithProperties.js
medou-boushab: I couldn't because I don't have this in intellij

fviale:
In general, very good, but I observe that Cut / Paste does not preserve task dependencies (tried with a replicate block)

medou-boushab:
> In general, very good, but I observe that Cut / Paste does not preserve task dependencies (tried with a replicate block)

I'm trying to fix this.

fviale:
@medou-boushab  Dependencies  are now preserved when we cut/paste, but not replicate script when pasting a replicate block. Other control structures are preserved (If, Loop).
Additionnally, undo does not work properly when we Cut.

PR #689, opened 2019-07-30
Add task property: fork

tao-xinxiu:
Add task property: fork

- add checkbox for the task parameter "Fork".
- move the checkbox "Run as me" under the tab "Fork Environment" with the checkbox "Fork".
- disable editing the related "Fork Environment" elements depending on the server global configuration and task "Fork", "Run as me" value.
	- if globally configured "fork", task is always "fork"
	- if globally configured "non-fork", task is always "non-fork"
	- if globally configured "runAsMe", task is always "fork"
	- if task "non-fork" (fork set to false), it's always "non-runAsMe", and "Fork Environment" elements are not editable and not taken effect.

Note, the client store global server properties in sessionStorage, which means they are only reloaded from the server when a new page session is initiated. (i.e., open a new tab, or re-open a closed the browser)


app/scripts/proactive/model/Task.js
```diff
@@ -347,6 +340,25 @@ define(
...
+                        "data-tab-help": "Fork environment is a new customisable JVM started to only run the task it belongs to. Also specify how to start this JVM, like in a Docker container for example.",
+                        'placeholder': '@attributes->fork',
+                        "data-help": 'Executes the task in a forked JVM, when it is false, all the other fork environment configurations are not taken into account.'
+                    }
+                },
+                "Run as me": {
+                    type: "Checkbox",
+                    fieldAttrs: {
+                        'placeholder': '@attributes->runAsMe',
+                        "data-help": 'Executes the task under your system account.'
```
fviale: in the help, you should add that this setting implies as well fork

tao-xinxiu: Done. Added: ", it also implies the task is executed in a forked JVM."

PR #692, opened 2019-08-09
feat: task retry delay

tao-xinxiu:
Add the "Delay Before Retry Task" property (format `hh:mm:ss`) in job and task error management.


tao-xinxiu:
retest this please

app/scripts/proactive/model/Job.js
```diff
@@ -218,6 +218,13 @@ define(
             "data-help": "Defines whether tasks that have to be restarted will restart on an other computing node."
           },
           options: ["anywhere", "elsewhere"]
+        },
+        "Delay Before Retry Task (hh:mm:ss)": {
+            type: "Text",
+            fieldAttrs: {
+                'placeholder': '@attributes->taskRetryDelay',
+                "data-help": 'Specifies how long to wait before restart the task in error. <br/><br/>Format is the following:<br/><br/>5 means 5 seconds<br/><br/>10:5 means 10 minutes 5 seconds<br/><br/>1:02:03 is 1 hour 2 minutes and 3 seconds.'
```
mcompastie: The data-help attribute should be: `This parameter specifies gow long to wait before restarting the task in error`

tao-xinxiu: The word `restarting` in the data-help cause the test fail (maybe because the word is too long). I haven't found a solution, so I leave the data-help as it is for now.

PR #694, opened 2019-08-22
feat: add workflow variable type `PA:HIDDEN`

tao-xinxiu:
When the workflow variable has the type PA:HIDDEN, its value is hidden during input.

### DIFF
--- a/app/scripts/proactive/templates/job-variable-template.html
+++ b/app/scripts/proactive/templates/job-variable-template.html
@@ -1,5 +1,5 @@
 <center>
-<% _.each(jobGenericInfos, function(gi) { 
+<% _.each(jobGenericInfos, function(gi) {
     		var label = gi["Property Name"];
     		if (label == "workflow.icon" )  %>
     			<img src="<%=gi["Property Value"]%>" style="width:100px;height:100px;">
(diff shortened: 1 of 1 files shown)


app/scripts/proactive/templates/job-variable-template.html
```diff
@@ -53,6 +53,9 @@ <h4 class="modal-title"><%- jobName %> </h4>
                     <label class="radio-inline"><input type="radio" name="<%- variable.Name %>" value="true"/>TRUE</label>
                     <label class="radio-inline"><input type="radio" name="<%- variable.Name %>" value="false" checked/>FALSE</label>
                 </div>
+            <% } else if (variable.hasOwnProperty("Model") && variable.Model.toLowerCase().indexOf('pa:hidden')!=-1) {  %>
+                <input class="variableValue textareavalues" style="border-radius: 4px; border: 1px solid #ccc; vertical-align: middle; color: #555;" name="<%- variable.Name %>" type="password" data-variable-model="<%- variable.Model %>" id="<%- key %>" value="<%- variable.Value %>" />
+                <label class="Extra" data-toggle="tooltip" title="<%- variable.Model %>"><%- variable.hasOwnProperty("Model") ? ((variable.Model.length >= 20) ? variable.Model.substring(0, 20) + " ..." : variable.Model):""  %></label>
```
Mykhailenko: sorry. i have to ask. what is this `20` constant means here?

tao-xinxiu: I copied it from the code for all the other types of variable. It shows "..." when the variable Model definition is too long. (But I don't know why it was the constant `20`).

PR #702, opened 2019-09-16
fix [jobVar]: init jobVariableView.info to avoid null jobInfo

tao-xinxiu:


app/scripts/proactive/view/JobVariableView.js
```diff
@@ -12,7 +12,7 @@ define(
 
         template: _.template(jobVariableTemplate),
 
-        model: undefined,
+        model: {'jobVariables': {}, 'jobName':'', 'jobProjectName':'', 'jobDescription':'', 'jobDocumentation':'', 'jobGenericInfos':'', 'errorMessage':'', 'infoMessage' :''},
```
amouhoub: If the error was only because model was undefined then you could simply do `model:{}`
If the error was raised because some code tried to access a property of model (jobVariables, etc ) then it's fine like you did it.

tao-xinxiu: Yes, the error was because `jobProjectName` and `jobDescription` could be undefined, while they are required in the template html.

tao-xinxiu:
> I am not sure jobGenericInfos should be initialized to empty string. Could you double check ?

Done. Changed to an array.

PR #707, opened 2019-10-14
Bug in Studio: Import as current wf doesn't contain all nodes.

medou-boushab:


app/scripts/proactive/model/SchemaModel.js
```diff
@@ -79,12 +79,24 @@ define(
                             if (!Array.isArray(value)) {
                                 value = [value];
                             }
-                            $.each(value, function(i, v) {
-                                var listElemValue = that.getListElementValue(that.schema[prop], v)
-                                if (listElemValue) {
-                                    newElements.push(listElemValue)
-                                }
-                            })
+                            if(this.schema[prop].itemType === 'NestedModel' && value[0].script.length > 1 ){
```
fviale: I don't think you can assume here that:
value[0].script.length exists all the time. In most element of the model, value[0].script will be undefined. Raising an error

PR #714, opened 2019-11-19
Add automatic encryption for PA:HIDDEN variables

fviale:
- When creating or editing a PA:HIDDEN variable in the workflow, the server response contains the data in encrypted format. Use this value to store it in the workflow instead of clear text data.
 - like often, backbone-forms did not provide the necessary feature, so I resolved on modifying backbone-forms to give access to the current form when validating a list entry. This allows the validation to modify the value when the form is submitted.

### DIFF
--- a/app/scripts/proactive/model/Job.js
+++ b/app/scripts/proactive/model/Job.js
@@ -259,7 +259,7 @@ define(
 
         var that = this;
         this.schema.Variables.subSchema.Model.validators = [
-          function checkVariableValue(value, formValues) {
+          function checkVariableValue(value, formValues, form) {
             if (formValues.Model.length > 0) {
               if (StudioApp.isWorkflowOpen()) {
                 that.updateVariable(formValues);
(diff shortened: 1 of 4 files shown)


app/scripts/proactive/model/Job.js
```diff
@@ -270,6 +270,12 @@ define(
...
                   return err;
+                } else if (formValues.Model === "PA:HIDDEN" && formValues.Value.trim().length > 0) {
+                    formValues.Value = validationData.updatedVariables[formValues.Name];
+                    if (form) {
+                        form.setValue(formValues);
+                    }
+                    delete that.attributes.BackupVariables;
                 } else {
                     delete that.attributes.BackupVariables;
                 }
```
MaelAudren: You can replace 
```
[...]
     delete that.attributes.BackupVariables;
 } else {
     delete that.attributes.BackupVariables;
}
```
by
```
[...]
}
delete that.attributes.BackupVariables;
```

app/scripts/proactive/model/Task.js
```diff
@@ -457,6 +457,12 @@ define(
...
                                     return err;
+                                } else if (formValues.Model === "PA:HIDDEN" && formValues.Value.trim().length > 0) {
+                                    formValues.Value = validationData.updatedVariables[that.attributes["Task Name"] + ":" + formValues.Name];
+                                    if (form) {
+                                        form.setValue(formValues);
+                                    }
+                                    delete that.attributes.BackupVariables;
                                 } else {
                                     delete that.attributes.BackupVariables;
                                 }
```
MaelAudren: same remark as upper

app/scripts/proactive/model/Job.js
fviale: unfortunately not, because the PR does not show the if case which does not call delete that.attributes.BackupVariables;

fviale: actually you're right, the if case calls return immediately

PR #716, opened 2019-12-03
Add tooltips to show task descriptions in canvas

medou-boushab:


app/scripts/proactive/view/TaskView.js
```diff
@@ -86,10 +86,18 @@ define(
             this.model.on("change:Generic Information", this.updateIcon, this);
 
             this.model.on("invalid", this.setInvalid, this);
-                    
-            this.element = $('<div class="task"><a class="task-name"><img src="'
-                + iconPath + '" width="20px">&nbsp;<span class="name">'
-                + this.model.get("Task Name") + '</span></a></div>');
+            if(this.model.get("Description")){
+                this.element = $('<div class="task"><a class="task-name"><img src="'
```
amouhoub: To avoid duplication define a variable with the description or the "This task has no description" and use it in the template.

PR #720, opened 2019-12-05
Studio:: Add tooltips to show task descriptions in canvas

medou-boushab:


medou-boushab:
It has many commits not related to the title.

Issue #717, opened 2019-12-03
Workflow name is not encoded when querying the catalog


MaelAudren:
The request send by the studio to the Catalog is 
`GET http://localhost:8080/catalog/buckets/basic-examples/resources/test`
when it should be 
`GET http://localhost:8080/catalog/buckets/basic-examples/resources/test%231`

Encoding the workflow name in the URL should fix the issue.

Reproducing steps:
1/ Push to the catalog a workflow named `test#1`
2/ Create a new workflow
3/ Click on Catalog -> Get a workflow from the catalog
4/ Select the bucket where you push `test#1` then select `test#1`
Issues: 
- The revision are not displayed
- The workflow description match the previous workflow
- Using the button `Append to the current workflow` return a 500 Error

medou-boushab:
fixed!

PR #728, opened 2019-12-24
Studio::: Workflow name is not encoded when querying the catalog

medou-boushab:
https://github.com/ow2-proactive/studio/issues/717


medou-boushab:
@amouhoub Not needed because we fallow some rules in catalog.

PR #730, opened 2019-12-30
On all connection pages, add the Pictograms and (in bigger) the name and description

medou-boushab:
On all connection pages, add the Pictograms and (in bigger) the name and description as in page: https://trydev.activeeon.com/


app/scripts/proactive/templates/login-template.html
```diff
@@ -3,33 +3,33 @@
...
-                        <select id="login-mode" name="login-mode" class="controls font-size-11">
+                        <select id="login-mode" name="login-mode" class="controls font-size-11" style="width:170.667px">
                             <option value="basic">Basic</option>
                             <option value="credentials">Credentials</option>
                         </select>
                     </div>
                     <div id="login-basic">
                         <div class="form-group">
-                            <label class="control-label" for="user">User: </label>
+                            <label class="control-label" for="user">Username: </label>
```
aminelouati: Why you changed user to userneme here and not everywhere ?

medou-boushab: We have one login page ? where do you see user ?

Issue #675, opened 2019-06-07
Replacing "if" task in the "if" workflow does not work


MaelAudren:
Steps to reproduce:
1/ Create a new workflow
2/ Create a new task
3/ Move the if arrow to the new task
4/ Remove the original if task

By exporting the task, you can see that the controlFlow tag is broken. `target` attribute is not modified when `continuation` attribute target the new task.

PR #737, opened 2020-01-16
fix issue #675 for updating if branch

tao-xinxiu:
When changed if/else branch connected target task in studio, controlFlow target (if/else/continuation) should be correctly updated in the workflow definition. 

Previously, it always change the "continuation" target to the new associated task, even when it's actually "if" or "else" connection updated in front-end. So it caused inconsistency between workflow diagram and its model definition.

Fix:
- retrieve the updated branch from the connection label, then call the corresponding setif/setelse/setcontinuation function to update task model.
- setif should only initialize the script model to `new FlowScript()` when the control flow "if" is newly set, so that updating if branch target won't erase existing controlFlow script.

### DIFF
--- a/app/scripts/proactive/model/Task.js
+++ b/app/scripts/proactive/model/Task.js
@@ -632,16 +632,6 @@ define(
                 }
             },
             setControlFlow: function (controlFlowType, task) {
-                if (controlFlowType == 'if') {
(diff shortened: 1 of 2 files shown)


tao-xinxiu:
retest this please

PR #744, opened 2020-01-30
Structure tasks menu

aminelouati:
In this PR we:
1. Structure the Tasks menu by adding three sections: OS, Languages and Containers
2. Delete useless templates 
3. Delete useless template icons


fviale:
test failed (could be because you removed some workflows)

aminelouati:
re-test this please

aminelouati:
I removed workflows of manuals and controls menus because they are moved to the Catalog. I also removed script_cron.xml and Ldap_query.xml templates moved to other menus and an old template native.xml. Tests failed because of instability.

medou-boushab:
do you  need this extension ".gif"?  There is If.gif and If.png

PR #750, opened 2020-03-05
Improve singularity script

fviale:


fviale:
Retest this please

app/scripts/proactive/config.js
```diff
@@ -273,91 +273,128 @@ println "DOCKER COMMAND : " + forkEnvironment.getPreJavaCommand()
...
+        default:
+            isWindows = false;
+    }
 
-containerName = "java"
-containerUrl = "docker://" + containerName
+    process = "singularity --version".execute()
+    majorVersion = 0
+    if (process.waitFor() == 0) {
+        version = process.text
```
nebilbenmabrouk: we can display singularity version by adding:
println process.text OR println version

fviale: right, just added it

PR #754, opened 2020-04-21
Add file browser modal for showing dataspace files

tao-xinxiu:
For the variable PA:GLOBAL_FILE and PA:USER_FILE, add the modal to browse the corresponding data space files (support upload, delete file, create new folder), and select a data space file as the variable value.

### DIFF
--- a/app/scripts/proactive/templates/job-variable-template.html
+++ b/app/scripts/proactive/templates/job-variable-template.html
@@ -122,7 +122,7 @@
             <% } else if (model && model.toLowerCase().indexOf('pa:global_file')!=-1) {  %>
                 <textarea class="variableValue textareavalues submitVariableValue" rows=1 name="<%- variable.Name %>" data-variable-model="<%- model %>" id="<%- key %>" ><%- variable.Value %></textarea>
                 <label class="short" data-toggle="tooltip" title="<%- model %>"><%- model ? ((model.length >= 20) ? model.substring(0, 20) + " ..." : model):""  %></label>
-                <button id="var-globalfile-button" type="button" class="btn fas fa-link" style="background-color:Transparent"></button>
+                <button id="var-globalfile-button" value="<%- variable.Name %>" type="button" class="btn fas fa-link" style="background-color:Transparent"></button>
             <% } else {  %>
                 <textarea class="variableValue textareavalues submitVariableValue" rows=1 name="<%- key %>" data-variable-model="<%- model %>" id="<%- key %>" ><%- variable.Value %></textarea>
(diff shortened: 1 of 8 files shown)


app/scripts/proactive/templates/job-variable-template.html
```diff
@@ -118,7 +118,15 @@ <h4 class="modal-title"><%- jobName %> </h4>
...
+            <% } else if (model && model.toLowerCase().indexOf('pa:global_file')!=-1) {  %>
+                <textarea class="variableValue textareavalues submitVariableValue" rows=1 name="<%- variable.Name %>" data-variable-model="<%- model %>" id="<%- key %>" ><%- variable.Value %></textarea>
+                <label class="short" data-toggle="tooltip" title="<%- model %>"><%- model ? ((model.length >= 20) ? model.substring(0, 20) + " ..." : model):""  %></label>
+                <button title="Browse global space files" value="<%- key %>" class="var-globalfile-button btn far fa-folder-open" type="button" style="background-color:Transparent"></button>
+            <% } else if (model && model.toLowerCase().indexOf('pa:user_file')!=-1) {  %>
+                <textarea class="variableValue textareavalues submitVariableValue" rows=1 name="<%- variable.Name %>" data-variable-model="<%- model %>" id="<%- key %>" ><%- variable.Value %></textarea>
+                <label class="short" data-toggle="tooltip" title="<%- model %>"><%- model ? ((model.length >= 20) ? model.substring(0, 20) + " ..." : model):""  %></label>
```
MaelAudren, line 128: Can't you create a method for `model ? ((model.length >= 20) ? model.substring(0, 20) + " ..." : model):""` instead of duplicate it ?

app/scripts/proactive/view/FileBrowserView.js
```diff
@@ -0,0 +1,228 @@
...
+                return;
+            }
+            var selectedFilePath = selectedElement.attr('value');
+            var result;
+            if(selectedElement.hasClass("file-browser-dir")) {
+                result = confirm(`Are you sure you want to permanently delete the folder "${selectedFilePath}" and all the files in it ?`);
+            } else {
+                result = confirm(`Are you sure you want to permanently delete the file "${selectedFilePath}" ?`);
+            }
+            if (result) {
```
MaelAudren, line 195: You can do
```
if(...){
    question=`Are you ...`
}else{
    question=`Are you ...`
}
if (confirm(question))
```
It avoid to duplidate the confirm call method.

app/scripts/proactive/templates/job-variable-template.html
tao-xinxiu: Actually I found that it's only necessary for the models like DATETIME which could be too long, so I'm removing those who don't need this expression.

PR #756, opened 2020-05-06
Add the possibility to open the workflow from the scheduler

mboussaa:


app/scripts/proactive/view/dom.js
```diff
@@ -279,6 +279,17 @@ define(
...
         }
 
+        function getWorkflowFromScheduler(url, successCallback) {
+            $.ajax({
+                url: url,
+                type: 'GET'
+            }).success(function (response) {
+                successCallback(response);
+            }).error(function (response) {
+                StudioClient.alert('Error', 'Error importing selected Workflow for the Scheduler: ' + JSON.stringify(response), 'error');
```
medou-boushab, line 289: from the Scheduler ?

app/scripts/proactive/app.js
```diff
@@ -205,6 +205,16 @@ define(
...
                 dom.open_catalog_workflow();
             });
+        },
+        openWorkflowFromScheduler : function(jobId) {
+            var that = this;
+            var sessionId = localStorage['pa.session'];
+
+            var url = '/scheduler/portal/downloadjobxml?jobId='+jobId+'&sessionId='+sessionId;
+            dom.getWorkflowFromScheduler(url, function (response) {
+                that.xmlToImport = new XMLSerializer().serializeToString(response);
```
MaelAudren, line 215: why using `that` instead of `this` ?
If I understand well xmlToImport contains the response String. Where is it used ?

mboussaa: I used seamless mechanism as catalog. `xmlToImport` is used by the import function to open the loaded `xml`.

app/scripts/proactive/view/dom.js
mboussaa: done

app/scripts/proactive/app.js
mboussaa: @MaelAudren you can approve pls?

mboussaa:
re-test this please

PR #760, opened 2020-05-13
Cursor settings

medou-boushab:
- Auto focus to the new folder input box
- Use / for the root instead of \

### DIFF
--- a/app/scripts/proactive/templates/file-browser-template.html
+++ b/app/scripts/proactive/templates/file-browser-template.html
@@ -24,7 +24,7 @@
                 <span class="font-bold">
                     Location:
                 </span>
-                <span class="current-sub-path"><a href="javascript:void(0)" value="">\</a></span>
+                <span class="current-sub-path"><a href="javascript:void(0)" value="">/</a></span>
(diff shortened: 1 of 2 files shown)


app/scripts/proactive/templates/file-browser-template.html
```diff
@@ -24,7 +24,7 @@ <h3><%=locationDescription%> Files</h3>
                 <span class="font-bold">
                     Location:
                 </span>
-                <span class="current-sub-path"><a href="javascript:void(0)" value="">\</a></span>
+                <span class="current-sub-path"><a href="javascript:void(0)" value="">/</a></span>
```
ShatalovYaroslav, line 27: Is this corresponds to Path Separator?
If yes, path separator depends on OS platform.
"\" in Windows, "/" in Unix

https://stackoverflow.com/questions/125813/how-to-determine-the-os-path-separator-in-javascript
from other side here it says "you can always use / as a path separator"

So please check the modifications on both OS.

medou-boushab: It's not a real Path, it'll be only on the model, allows us to navigate in folder

PR #765, opened 2020-05-20
change file browser modal size

tao-xinxiu:


app/scripts/proactive/templates/file-browser-template.html
```diff
@@ -1,4 +1,4 @@
-<div class="modal-dialog">
+<div class="modal-dialog" style="width:60vw">
```
amouhoub, line 1: why not add this to the class `modal-style` in css if it applies to all modal-class divs ?

tao-xinxiu: This one is specific to the file browser modal now. (It requires a larger width as it has more info in each line)

tao-xinxiu:
retest this please

app/scripts/proactive/templates/file-browser-template.html
medou-boushab: can use bootstrap class .modal-sm or .modal-lg ?

tao-xinxiu:
In studio, we don't have same bootstrap version, so can't use modal-lg for now.

app/scripts/proactive/templates/file-browser-template.html
medou-boushab: https://getbootstrap.com/docs/4.2/components/modal/

medou-boushab:
> In studio, we don't have same bootstrap version, so can't use modal-lg for now.

ok

app/scripts/proactive/templates/file-browser-template.html
```diff
@@ -34,7 +34,7 @@ <h5>Browse, Upload files in <%=locationDescription%>, Select one file as the var
                 </ol>
             </div>
             <br>
-            <div id="files-container" style="overflow-x: hidden; overflow-y: scroll; height: 270px; overflow-wrap: break-word;">
+            <div id="files-container" style="overflow-x: hidden; overflow-y: scroll; height: 30em; overflow-wrap: break-word;">
```
fviale, line 37: I think it would be best to use vh instead of em, it can this way adapt to different screens. @medou-boushab do you confirm ?

medou-boushab: @tao-xinxiu  I think in studio we can use max-height instead of height, this will reduce the empty space

PR #771, opened 2020-06-05
file browser modal: add support of download file/folder

tao-xinxiu:


app/scripts/proactive/view/FileBrowserView.js
```diff
@@ -272,7 +273,40 @@ define(
             });
         },
 
-        deleteFile: function(event) {
+        downloadFile: function() {
+            var selectedElement=$("#files-tbody  tr.selected").children().first();
+            if (selectedElement.length == 0) {
```
medou-boushab, line 278: Does it mean you check if there is selected className? if yes I think checking the  length of $("#files-tbody  td.active") will make the code more readable

tao-xinxiu: Since it's also used to get the filename of the selected row.

medou-boushab: yes but you can do this after, it's better than chaining children() and first() of an element that might not exist

tao-xinxiu: You're right. Thanks for the great suggestion. But since currently selected class is assigned only to tr, it will require quite a lot changes to all the places where have used selected file. I test the current version, not get error when none selected. So I propose to do it after the release.

medou-boushab: ok

PR #785, opened 2020-10-05
Studio:: variables substitution

medou-boushab:
The objective is to consider the value of $B as the PA model of A as well in submission modals and show its PA_MODEL name or UI controls (boolean switches, etc)

### DIFF
--- a/app/scripts/proactive/templates/job-variable-template.html
+++ b/app/scripts/proactive/templates/job-variable-template.html
@@ -64,10 +64,10 @@
             <label for="<%- key %>" class="Name" title="<%- key %>" data-toggle="tooltip"><%- (key.split(":").length == 2 ? key.split(":")[1] : key) %></label>
             <%
                var model = null;
-               var referenceType = null;
+               var resolvedModel = null;
                if (variable.hasOwnProperty("Model")) {
                    model = variable.Model;
-                   referenceType = variable.referenceType;
+                   resolvedModel = variable.resolvedModel;
                    if (model && model.toLowerCase().indexOf('pa:model_from_url') != -1) {
                         // fetch the specified urls and replace the current model by the result
                         var indexBegin = model.indexOf('(');
@@ -106,7 +106,7 @@
                                 <option><%- option %></option>
                     <% } %>
                 </select>
-            <% } else if ((model && model.toLowerCase().indexOf('pa:boolean')!=-1) || (referenceType  && referenceType.toLowerCase().indexOf('pa:boolean')!=-1) ) {  %>
+            <% } else if ((model && model.toLowerCase().indexOf('pa:boolean')!=-1) || (resolvedModel  && resolvedModel.toLowerCase().indexOf('pa:boolean')!=-1) ) {  %>
                 <div class="variableValue radio-div" id="<%- key %>" onchange="$('#check-button').click()" data-variable-model="<%- model %>">
                 <% if (variable.Value && (variable.Value.toLowerCase() === 'true' || Number(variable.Value) > 0)) %>
                     <label class="radio-inline"><input type="radio" name="<%- key %>" value="true" checked/>TRUE</label>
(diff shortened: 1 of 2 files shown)


app/scripts/proactive/templates/job-variable-template.html
```diff
@@ -136,10 +138,13 @@ <h4 class="modal-title"><%- jobName %> </h4>
...
                 <label class="short" data-toggle="tooltip" title="<%- model %>"><%- shortenModel %></label>
                 <button title="Browse, Upload User Space files, and Select one folder" value="<%- key %>" class="var-userfolder-button btn far fa-folder-open" type="button" style="background-color:Transparent"></button>
-            <% } else {  %>
-                <textarea class="variableValue textareavalues submitVariableValue" rows=1 name="<%- key %>" data-variable-model="<%- model %>" id="<%- key %>" ><%- variable.Value %></textarea>
-                <label class="Extra" data-toggle="tooltip" title="<%- model %>"><%- shortenModel %></label>
-            <% }  %>
+            <% } else { %>
+                <textarea class="variableValue textareavalues submitVariableValue" rows=1 name="<%- variable.Name %>" data-variable-model="<%- model %>" id="<%- key %>" ><%- variable.Value %></textarea>
+                <label class="short" data-toggle="tooltip" title="<%- model %>"><%- referenceType %></label>
+                <% if( referenceType === "PA:GLOBAL_FILE" ) { %>
```
tao-xinxiu, line 144: Why only for "PA:GLOBAL_FILE" but not "PA:USER_FILE", "PA:USER_FOLDER", and "PA:GLOBAL_FOLDER" ?

app/scripts/proactive/view/dom.js
```diff
@@ -428,6 +428,11 @@ define(
                 return;
             }
 
+            for( var key in jobVariables){
+                if(jobVariables[key].Model.indexOf('$') !== -1){
```
fviale, line 432: This will clash too much with models which can contain $ (for example regexp).

The correct way is to build a set or list of possible variable references by iterating on the keys first and adding in this set/list, for each key:

$key
${key}

app/scripts/proactive/view/dom.js
```diff
@@ -428,6 +428,11 @@ define(
...
+                    jobVariables[key].referenceType = jobVariables[jobVariables[key].Model.split('$')[1]].Value;
```
fviale, line 433: I would name this variable jobVariables[key].resolvedModel.
And I would do:
- jobVariables[key].resolvedModel=jobVariables[key].Model
- for each element pattern of the map built above:
   jobVariables[key].resolvedModel = jobVariables[key].resolvedModel.replace(pattern, replacement)

app/scripts/proactive/templates/job-variable-template.html
tao-xinxiu: It seems to me that dealing with referenceType is similar to mdoel. It could be put in the same place.

app/scripts/proactive/templates/job-variable-template.html
```diff
@@ -139,11 +139,27 @@ <h4 class="modal-title"><%- jobName %> </h4>
...
-                <label class="short" data-toggle="tooltip" title="<%- model %>"><%- referenceType %></label>
-                <% if( referenceType === "PA:GLOBAL_FILE" ) { %>
+                <% } else  { %>
+                <input class="variableValue textareavalues submitVariableValue" name="<%- key %>" type="password" data-variable-model="<%- model %>" id="<%- key %>" value="<%- variable.Value %>" autocomplete="new-password" />
+                <% } %>
+                <label class="short" data-toggle="tooltip" title="<%- model %>"><%- resolvedModel %></label>
+                <% if( resolvedModel && resolvedModel.toLowerCase().indexOf('pa:global_file') != -1 ) { %>
                 <button title="Browse Global Space files, Upload, and Select one file" value="<%- key %>" class="var-globalfile-button btn far fa-folder-open" type="button" style="background-color:Transparent"></button>
                 <% } %>
+                <% if(resolvedModel && resolvedModel.toLowerCase().indexOf('pa:user_file') != -1 ) { %>
```
tao-xinxiu, line 151: Why not combining these conditions with the previous ones for model like `if( (model && model.toLowerCase().indexOf('HANDLING_MODEL')!=-1) || (resolvedModel  && resolvedModel.toLowerCase().indexOf('HANDLING_MODEL')!=-1) )` to reduce the duplicated code for a specific model's display.
p.s. I think you can also create a function to check whether `model` or `resolvedModel` matches a specific handling model.

mbenguig: Maybe you could have added some comments to explain the conditions

mbenguig (approved):
Just a small comment.

tao-xinxiu (approved):

fviale:
I tested locally and it works fine, great work !

fviale (approved):

PR #787, opened 2020-11-30
Docker forkenv: use volumes-from when a specific java property is defined

fviale:


app/scripts/proactive/config.js
```diff
@@ -241,23 +242,37 @@ switch (family) {
...
+}
 // Prepare working directory (For Dataspaces and serialized task file)
 workspaceHost = localspace
 workspaceContainer = (isWindows ? forkEnvironment.convertToLinuxPath(workspaceHost) : workspaceHost)
-cmd.add("-v")
-cmd.add(workspaceHost + ":" + workspaceContainer)
+if (!isPANodeInContainer) {
+    cmd.add("-v")
+    cmd.add(workspaceHost + ":" + workspaceContainer)
+}
```
mcompastie, line 266: /!\ When the ProActive agent run in a docker container & interface the docker daemon from the host, it is necessary to take into account both of these process do not have the same file system. In @mboussaa work, communication between those two filesystems was achieved by sharing a volume mounted in `/tmp/tmp_node/tmp/:/tmp/
`. Besides, it is needed for having  common localspace folder, required for forkenv.

In the forkenv  prototype, this feature was implemented w/ the following code:
```
# Prepare working directory (For Dataspaces and serialized task file) 
workspaceHost = localspace 
workspaceContainer = localspace
workspaceVolume = '-v /tmp/tmp_node'+localspace +':'+localspace+' '
print 'workspaceVolume=' + workspaceVolume
```

mcompastie (changes requested):
One question regarding local-space management between forked and non-forked env to be considered.

app/scripts/proactive/config.js
mboussaa: @mcompastie 
I am not doing `/tmp/tmp_node/tmp/:/tmp/` but I am only mounting: `/tmp` (which will be configured later on at node startup to be `/tmp/node` to avoid mounting all `tmp`)

if we omit this condition, we will have something like this:
`-v, /tmp/PA_JVM1178115964/local__localhost__0_4/38t0/786017111:/tmp/PA_JVM1178115964/local__localhost__0_4/38t0/786017111`

the /tmp being provided by the volumes-from, there is no need to have this mounting point when using docker in docker

mcompastie: Ok, I though --volume-from had a cloning behavior. In that case, I understand this is not strictly necessary to manually share container:///tmp/ with the host.

mcompastie (approved):

PR #790, opened 2021-01-04
Fix bug:: avoid unnecessary initialize values

medou-boushab:


app/scripts/proactive/view/dom.js
```diff
@@ -550,10 +550,13 @@ define(
                         inputVariables[key].resolvedModel = validationData.updatedModels[key];
                         //Variables substitution:: When one changes the reference model we should initialize the Value and not show error message
                         var jobVariables = studioApp.views.jobVariableView.model.jobVariables;
-                        if(jobVariables[key].resolvedModel !== inputVariables[key].resolvedModel ){// compare old resolvedModel with the new resolvedModel
-                            inputVariables[key]['Value'] = "";
-                            validationData.errorMessage = "";
+                        if(jobVariables[key].resolvedModel){
+                            if(jobVariables[key].resolvedModel !== inputVariables[key].resolvedModel ){// compare old resolvedModel with the new resolvedModel
```
amouhoub, line 554: since there is only an if inside the first if => put them in the same condition with &&

```if ( jobVariables[key].resolvedModel && jobVariables[key].resolvedModel !== inputVariables[key].resolvedModel)```

amouhoub (changes requested):

amouhoub (approved):

fviale (approved):

mboussaa (approved):

PR #791, opened 2021-02-05
File Browser : add file permissions and hidden files

fviale:
- add file permissions in the form "rwx" to file list
 - add check box to hide/show files or directories starting with .
 - improve error messages in case of unauthorized file or folder operations

### DIFF
--- a/app/scripts/proactive/templates/file-browser-template.html
+++ b/app/scripts/proactive/templates/file-browser-template.html
@@ -28,6 +28,9 @@
                 <a id="delete-file-btn" href="javascript:void(0)" class="pointer pa-tool-btn" title="Delete the selected file or folder">
                     <i class="fas fa-trash"></i>
                 </a>
+                <div class="pull-right" id="pull-right-hidden-files-checkbox">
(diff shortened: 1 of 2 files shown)


tao-xinxiu (approved):
LGTM

app/scripts/proactive/templates/file-browser-template.html
```diff
@@ -47,27 +50,36 @@ <h5>
...
+                        <td> <%- directories[index].rights %> </td>
                         <td> </td>
                         <td> Folder </td>
                         <td> <%- directories[index].modified %> </td>
                     </tr>
-                    <% } %>
-                    <% for (var index in files) { %>
+                    <%   }
+                       } %>
+                    <% for (var index in files) {
```
MaelAudren, line 71: I don't think you need to close then open the `%>` and `<%` tokens here.

MaelAudren (approved):

app/scripts/proactive/templates/file-browser-template.html
fviale: yes, right

fviale:
Retest this please

PR #800, opened 2021-04-09
[outdated] Allow uploading after close dataspace file browser modal with showing progress bar

tao-xinxiu:
This PR is now **outdated**. Another design (showing uploading files in a dropdown list) have been decided for the Automation Dashboard. The similar design is expected to be done for the studio in the future.

This PR was used to:

- move the dataspace file uploading outside file browser modal controller
- continue uploading files after close file browser modal
- show the onging uploading progress bar in the toastr notification
- allow uploading multiple files at the same time
- add the dependency of lib toastr
studio_two_uploading
studio_two_uploaded


mboussaa:
@tao-xinxiu If you keep images in the PR you find issues after merging the PR. I suggest removing them. This happened to me only to studio PR before

tao-xinxiu:
> @tao-xinxiu If you keep images in the PR you find issues after merging the PR. I suggest removing them. This happened to me only to studio PR before

@mboussaa Thank you. I'll remove them before merging the PR.

medou-boushab (approved):

amouhoub:
@tao-xinxiu One general comment, non blocking, but you should use the pnotify (https://sciactive.com/pnotify/), the equivalent of toastr that we already use in the studio. (Workflow validation & submission messages)
We have to respect the actual graphical chart and components of the studio.
If it won't take you too much time, I strongly recommend using pnotify instead of toastr. I think it won't be too much to do as you have already implemented the logic behind. It's just the display module.

PR #832, opened 2022-01-27
Remember which tasks were toggled when the variables are validated.

fviale:
Also, add Main Variables label to empty groups

### DIFF
--- a/app/scripts/proactive/view/dom.js
+++ b/app/scripts/proactive/view/dom.js
@@ -437,6 +437,18 @@ define(
             }
         }
 
+        function getToggledTasks() {
+            var toggler = document.getElementsByClassName("caretUL");
+            var toggledTasks = [];
+            for (let i = 0; i < toggler.length; i++) {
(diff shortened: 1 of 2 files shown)


medou-boushab (approved):

app/scripts/proactive/view/dom.js
```diff
@@ -437,6 +437,19 @@ define(
             }
         }
 
+        function getToggledTasks() {
+            var toggler = document.getElementsByClassName("caretUL");
+            var i;
+            var toggledTasks = [];
+            for (i = 0; i < toggler.length; i++) {
```
tao-xinxiu, line 444: `i` could be declared in the loop like `for (let i = 0; i < cars.length; i++) {`

tao-xinxiu (approved):
LGTM, one tiny comment.

app/scripts/proactive/view/dom.js
fviale: ok, thanks for the trick :-)

PR #833, opened 2022-02-02
Improve multiple-lines 3rd-party-credential UX

tao-xinxiu:
- Improve UX when switching between the single-line and multiple-lines credential value. 
	- The credential value is saved when switching between the single-line and multiple-lines, and only single-line value is inputted.
	- Show the error message and erase the credential content when switching from inputted multi-lines cred to single-line. 
	- The error message is removed after the user inputted any character.
- Support auto-fill credential key to add/edit when the PA:CREDENTIAL variable value is filled.


fviale (approved):
LGTM

PR #836, opened 2022-02-21
add-filtering-capabilities-to-user-and-global-dataspace-file-browsers

oanaBiancaSchiopu:


app/scripts/proactive/view/FileBrowserView.js
```diff
@@ -113,7 +123,7 @@ define(
             }
             $.ajax({
                 url: that.dataspaceRestUrl + encodeURIComponent(pathname),
-                data: { "comp": "list" },
+                data: { "comp": "list", "includes": '*' + this.filterValue + '*'},
```
fviale, line 126: I would rather have the filterValue directly applied as includes. The default value of filterValue can be '*'

app/scripts/proactive/templates/file-browser-template.html
```diff
@@ -28,6 +28,7 @@ <h5>
                 <a id="delete-file-btn" href="javascript:void(0)" class="pointer pa-tool-btn" title="Delete the selected file or folder">
                     <i class="fas fa-trash"></i>
                 </a>
+                <label style="padding-left: 160px">Filter: <input type="text" id="filter-files"/> </label>
```
fviale, line 31: can you add a title attribute to the label "Filter files according to a glob pattern"

fviale (approved):
LGTM

app/scripts/proactive/templates/file-browser-template.html
```diff
@@ -28,6 +28,7 @@ <h5>
                 <a id="delete-file-btn" href="javascript:void(0)" class="pointer pa-tool-btn" title="Delete the selected file or folder">
                     <i class="fas fa-trash"></i>
                 </a>
+                <label title="Filter files according to a glob pattern" style="padding-left: 160px">Filter: <input type="text" value="*" id="filter-files"/> </label>
```
mklkun, line 31: glob or global?

fviale: it's called a glob pattern.
https://en.wikipedia.org/wiki/Glob_(programming)

mklkun: I learned something new. Thanks!

PR #850, opened 2022-04-12
catalog RBAC: disable publish button when the user has no write permission

tao-xinxiu:
- when publishing a new workflow or script, if the user is publishing a new revision, check the object rights, if the user is publishing a new object, check the bucket rights. Only when the user has the write or admin rights, the publish button is enabled. Otherwise, it's disabled with the corresponding error message in the tooltip.
- Add the owner and rights in the buckets tooltips, and add rights in the catalog objects tooltips.
- In the modal of publishing a script to the catalog, distinguish the UX between adding a new script and a new revision. 
	- When the user wants to publish a new revision, they just select the object.
	- When the user wants to create a new script object, they can input the new script name in a textbox (at top) in the objects list.
- Since adding a new script or a new revision is distinguished, adding a new revision now use the template  `catalog-publish-description.html` instead of  `catalog-publish-description-first.html` like publishing a workflow.
- Fix the bug when publishing a new revision of a script.

Screenshot when first open the modal to publish a script to catalog:
Screenshot from 2022-04-13 17-22-27

Screenshot for publishing a new revision of catalog object:
Screenshot from 2022-04-13 17-22-44

Screenshot for publishing a new catalolg object
Screenshot from 2022-04-13 17-22-59
:


tao-xinxiu:
retest this please

app/scripts/proactive/templates/catalog-publish.html
```diff
@@ -28,6 +28,10 @@
 						<div class="pull-left" id="catalog-publish-objects-title" style="font-weight: bold;padding-left: 15px;">Objects</div>
 					</legend>
 					<div style="height: 270px; overflow: auto;">
+						<div id="new-script-group" class="input-group w-100">
+							<input id="new-script-name" type="text" placeholder="new script name" class="form-control" style="display: table-cell">
+							<button id="add-new-script" class="btn btn-default"><i class="fa fa fa-plus"></i></button>
```
medou-boushab, line 33: Since it's the same action, you can use a className new-script-name for both (input and button)

I suggest to wrap them in form tag as we do in publish wf

app/scripts/proactive/view/CatalogPublishView.js
```diff
@@ -28,6 +28,8 @@ define(
         events: {
             'click #catalog-publish-buckets-table tr': 'selectBucket',
             'click #catalog-publish-objects-table tr': 'selectObject',
+            'change #new-script-name': 'addNewScript',
```
medou-boushab, line 31: see my comment above, not good to send requests for every change

medou-boushab (approved):

app/scripts/proactive/templates/catalog-publish.html
tao-xinxiu: Done. Thank you for the great suggestion.

app/scripts/proactive/templates/catalog-publish.html
```diff
@@ -28,6 +28,12 @@
 						<div class="pull-left" id="catalog-publish-objects-title" style="font-weight: bold;padding-left: 15px;">Objects</div>
 					</legend>
 					<div style="height: 270px; overflow: auto;">
+						<form class="container-flex w-100" id="new-script-form">
+							<div class="input-group w-100">
+								<input id="new-script-name" type="text" placeholder="new script name" class="form-control" style="display: table-cell">
+								<button id="add-new-script" class="btn btn-default"><i class="fa fa fa-plus"></i></button>
```
medou-boushab, line 34: do you still need `id="add-new-script"`  ?

you should add `type="submit"` to the button

oanaBiancaSchiopu (approved):

tao-xinxiu:
retest this please

PR #867, opened 2022-07-12
publish workflow: editable kind and correct default kind workflow/psa

tao-xinxiu:
- When publish a workflow, make its kind property editable. 
- When the workflow has the generic info `pca.states` and `pca.service.id` specified, change its default kind value to workflow/psa.


medou-boushab (approved):

fviale (approved):
LGTM

PR #878, opened 2022-11-02
Add new tabs for the task properties to be coherent as workflow properties

tao-xinxiu:
Add new tabs 'Task Variables' and 'Generic Info' for the task properties to be coherent as workflow properties.

### DIFF
--- a/app/scripts/proactive/model/Task.js
+++ b/app/scripts/proactive/model/Task.js
@@ -88,11 +88,12 @@ define(
                     confirmDelete: 'You are about to remove a variable.'
                 },
                 "Generic Info": {
+                    title: 'Task Generic Info',
                     type: 'List',
                     itemType: 'Object',
                     fieldAttrs: {
                         'placeholder': 'genericInformation->info',
-                        "data-tab": "Generic Info",
+                        "data-tab": "Task Generic Info",
                         "data-tab-help": 'Some extra information about your job often used to change the scheduling behavior for a job. E.g. NODE_ACCESS_TOKEN=rack1 will assign this task to a node with token \"rack1\".'
                     },
                     itemToString: Utils.inlineNameValue, itemTemplate: Utils.bigCrossTemplate,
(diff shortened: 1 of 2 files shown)


app/scripts/proactive/model/Task.js
```diff
@@ -58,13 +58,21 @@ define(
...
+                        "data-help": 'Indicates if you want to save the result of this task in the job result.'
+                    }
+                },
                 "Variables": {
                     type: 'List',
                     itemType: 'Object',
                     title: "Task Variables (Name, Value, Inherited)",
                     fieldAttrs: {
                         'placeholder': 'variables->variable',
+                        "data-tab": "Task Variables",
```
fviale, line 74: why not Variables ?

tao-xinxiu: Here is the tab name, I feel it might be better to clarify it is just the "Task Variables"?

tao-xinxiu:
retest this please

app/scripts/proactive/model/Task.js
fviale: maybe you should ask d.c. before merging it ?

app/scripts/proactive/model/Task.js
```diff
@@ -84,7 +92,8 @@ define(
                     itemType: 'Object',
                     fieldAttrs: {
                         'placeholder': 'genericInformation->info',
-                        "data-help": 'Some extra information about your job often used to change the scheduling behavior for a job. E.g. NODE_ACCESS_TOKEN=rack1 will assign this task to a node with token \"rack1\".'
+                        "data-tab": "Generic Info",
```
mboussaa, line 95: Task Generic Info 
or Generic Task Info 
To be checked with d.c as well

mboussaa: you can show the screenshot in Trello and tag him to have his pov of both remarks

mboussaa (approved):

PR #892, opened 2023-02-17
Update generic info after publishing the wf to the Catalog

medou-boushab:


oanaBiancaSchiopu (approved):

app/scripts/proactive/view/dom.js
```diff
@@ -1122,6 +1122,26 @@ define(
 
         $("#confirm-publication-to-catalog").click(function () {
             var studioApp = require('StudioApp');
+            var studioApp = require('StudioApp');
```
fviale, line 1125: no need to duplicate this line

fviale (approved):
LGTM, one comment

app/scripts/proactive/view/dom.js
medou-boushab: yes, I'll remove it, thanks

amouhoub (approved):

PR #895, opened 2023-03-08
set-submission-mode

oanaBiancaSchiopu:


app/scripts/proactive/view/dom.js
```diff
@@ -785,6 +785,12 @@ define(
             var jobDocumentation = studioApp.models.jobModel.get("Generic Info Documentation");
             var jobGenericInfos = studioApp.models.jobModel.get("Generic Info");
 
+            const submissionMode = {
```
amouhoub, line 788: Just leave a comment here that explicitly explains this, so that we can remember where this is in the future, in case we want to change how we do this...

amouhoub (approved):

PR #905, opened 2023-04-26
Fix bugs: copy to keyboard doesn't work and select task from pipelines

medou-boushab:


medou-boushab:
I won't wait for reviews because it's all about moving a part of the code from his scope to another.

PR #919, opened 2023-07-24
Fix pattern usage in object name filters

fviale:
- add encodeUriComponent to prevent url issues
 - remove % replacement


mboussaa (approved):

mklkun (approved):
LGTM (with my limited front-end knowledge)

PR #920, opened 2023-07-27
add-domain-to-login-form

oanaBiancaSchiopu:
image

### DIFF
--- a/app/scripts/proactive/rest/studio-client.js
+++ b/app/scripts/proactive/rest/studio-client.js
@@ -230,6 +230,27 @@ define(
                 });
             },
 
+            getDomains : function () {
+                var domains=[];
+                $.ajax({
+                    type: 'GET',
+                    url: config.restApiUrl + "/domains",
+                    async: false,
+                    success: function(data) {
+                        if (!data.includes("")) {
+                            domains.push("");
+                        }
+                        for(var i = 0; i < data.length ; i++){
+                            domains.push(data[i]);
+                        }
+                    },
+                    error: function(data) {
+                            console.log("Problems to retrieve domains" + data.responseText)
+                    }
+                });
+                return domains;
+            },
+
             getPortalsAuthorizations : function () {
                 var result = null;
                 $.ajax({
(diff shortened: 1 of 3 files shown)


app/scripts/proactive/rest/studio-client.js
```diff
@@ -327,6 +348,7 @@ define(
                         that.alert("Cannot submit the job", result.errorMessage, 'error');
                     } else if (result.id) {
                         that.alert("Job submitted", "<html></html><a href='/scheduler' target='_blank'>'" + result.readableName + "' submitted successfully (Id " + result.id + ")</a></html>", 'success');
+                        that.alert("Job submitted", "<html></html><a href='/automation-dashboard/#/job-info?jobid=" + result.id + "&tab=0' target='_blank'>'" + result.readableName + "' submitted successfully (Id " + result.id + ")</a></html>", 'success');
```
fviale, line 351: I believe we need to have both links inside the same alert message.

Also we should rewrite it in order to understand these two links.
e.g.
`"<label>" + result.readableName + "' submitted successfully (Id " + result.id + ")+ "</label><br>"` 
then append the links (starting with automation dashboard) with an inside text like Open in Workflow Execution, Open in Scheduler

app/scripts/proactive/rest/studio-client.js
```diff
@@ -230,6 +230,27 @@ define(
...
 
+            getDomains : function () {
+                var domains=[];
+                $.ajax({
+                    type: 'GET',
+                    url: config.restApiUrl + "/domains",
+                    async: false,
+                    success: function(data) {
+                        if (!data.includes("")) {
+                            domains.push("");
```
fviale, line 241: I believe this will create a list [""] when the list of domains is empty

app/scripts/proactive/rest/studio-client.js
```diff
@@ -230,6 +230,27 @@ define(
...
 
+            getDomains : function () {
+                var domains=[];
+                $.ajax({
+                    type: 'GET',
+                    url: config.restApiUrl + "/domains",
+                    async: false,
+                    success: function(data) {
+                        if (data.length > 0 && !data.includes("")) {
+                            domains.push("");
```
AlexFalessi, line 241: Why do you add an empty string at the end of the list if there is returned data ?

oanaBiancaSchiopu: I add an empty string at the begining of the list so that the user has the possibility to login using a domain or not

AlexFalessi: Ok thanks, that makes sense.

AlexFalessi (approved):

fviale (approved):

PR #926, opened 2023-09-13
show-error-label-when-server-is-unresponsive

oanaBiancaSchiopu:
image

### DIFF
--- a/app/scripts/proactive/rest/studio-client.js
+++ b/app/scripts/proactive/rest/studio-client.js
@@ -197,7 +197,11 @@ define(
                             }
                         },
                         error: function(response) {
-                            console.error("Unknown User Permission", response)
+                            console.error("Unknown User Permission", response);
+                            var connectionError = $('<div id="connection-error" class="btn-group pa-tools"><label style="color:red; font-weight:bold; font-size:15px; padding-left:4px; padding-top:4px;">The server cannot be reached. Check the server status and your network connection. </label></div>');
+                            if (!$("#connection-error").is(":visible")) {
+                                $("#tool-bar").append(connectionError);
+                            }
                         }
                     });
                 } else {
--- a/app/scripts/proactive/view/LogoutView.js
+++ b/app/scripts/proactive/view/LogoutView.js
@@ -52,6 +52,11 @@ define(
                 function tryToConnect() {
                     if (StudioClient.isLoggedIn()) {
                         isConnected()
+                    } else {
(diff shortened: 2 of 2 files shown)


app/scripts/proactive/view/LogoutView.js
```diff
@@ -52,6 +52,11 @@ define(
                 function tryToConnect() {
                     if (StudioClient.isLoggedIn()) {
                         isConnected()
+                    } else {
+                        var connectionError = $('<div id="connection-error" class="btn-group pa-tools"><label style="color:red; font-weight:bold; font-size:15px; padding-left:4px; padding-top:4px;">You have been disconnected from the server. Reload the page in order to log into the portal again. </label></div>');
```
medou-boushab, line 56: why adding this in logout view ?

oanaBiancaSchiopu: this is where the check is done if the user is still logged in:
 if (StudioClient.isLoggedIn()) {

app/scripts/proactive/rest/studio-client.js
```diff
@@ -197,7 +197,11 @@ define(
                             }
                         },
                         error: function(response) {
-                            console.error("Unknown User Permission", response)
+                            console.error("Unknown User Permission", response);
+                            var connectionError = $('<div id="connection-error" class="btn-group pa-tools"><label style="color:red; font-weight:bold; font-size:15px; padding-left:4px; padding-top:4px;">The server cannot be reached. Check the server status and your network connection. </label></div>');
```
medou-boushab, line 201: I think it's not the right place, you should move it to the ligne 183

app/scripts/proactive/view/LogoutView.js
medou-boushab: yes, this means that you will wait for the second check:
isLoggedIn returns localStorage['pa.session'] != null
so before removing the session id it doesn't mean you are still connected.

app/scripts/proactive/rest/studio-client.js
oanaBiancaSchiopu: if I put it to line 183, it won't be executed in the case when the server is not running

app/scripts/proactive/view/LogoutView.js
oanaBiancaSchiopu: where do you think this should be added?

medou-boushab (approved):

PR #932, opened 2023-10-17
Update variables dispplay: advanced orange, hidden grey, normal black

AlexFalessi:


mboussaa:
please remove the screenshot before merging, because it will cause an issue on jenkins. This happens only on studio PR for an unknow reason. Thanks @AlexFalessi

mboussaa (approved):

PR #948, opened 2024-01-17
add-account-info

oanaBiancaSchiopu:


mboussaa (approved):

mboussaa:
@oanaBiancaSchiopu could you please remove the screenshot before merge otherwise it will cause issues thanks

fviale:
@oanaBiancaSchiopu 

can you add more "air" to the popup window margins ?

I think it is done using the css padding properties

fviale:
also, the title "Account Info" should be centered

fviale:
finally, why is username so big ?

oanaBiancaSchiopu:
> finally, why is username so big ?

it's a h3 tag, I can make it smaller if you want

fviale (approved):

PR #966, opened 2024-04-23
FIX: remove empty & null properties from variables when added or edited

amouhoub:


app/scripts/proactive/view/dom.js
```diff
@@ -945,6 +945,8 @@ define(
             });
             var actualVariable = variablesNewOrder[actualVarIndex];
 
+            console.log(variablesNewOrder)
```
medou-boushab, line 948: remove log

medou-boushab (approved):

mboussaa (approved):

PR #969, opened 2024-05-14
Preserve Flow script if it exists when connecting loop to another task

AlexFalessi:
Check if FlowScript exists and create only if not.


fviale (approved):

medou-boushab (approved):

AlexFalessi:
retest this please

PR #980, opened 2024-07-04
Read user name and domain and prefill form when reconnect

AlexFalessi:


fviale (approved):

app/scripts/proactive/view/LoginView.js
```diff
@@ -210,10 +220,18 @@ define(
                     domainOptionString = domainOptionString + '</select></div>';
                     var domainOption = $(domainOptionString);
                     $("#login-basic").prepend(domainOption);
+
+                    if (domainName && domains.includes(domainName)) {
```
medou-boushab, line 224: don't you need to check if domains is not null ?

AlexFalessi: It can't `var domainName = '';`. 
Actually I could have put only `domains.includes(domainName)`

medou-boushab: You mean it can't be null and domainName is not empty string ?

medou-boushab (approved):

PR #983, opened 2024-07-19
Fix undesired alert in Workflow Variables View

amouhoub:


app/scripts/proactive/templates/workflow-variables-view-template.html
```diff
@@ -191,7 +191,7 @@ <h4 class="modal-title"><%- jobModel.Name %> </h4>
                         model = data;
                         },
                         error: function(xhr, ajaxOptions, thrownError) {
-                            alert("Could not fetch model from url '" + urlToFetch + "': " + xhr.responseText);
+                            console.error("Could not fetch model from url '" + urlToFetch + "': " + xhr.responseText);
```
fviale, line 194: is there a console.warn ?

mboussaa (approved):

AlexFalessi (approved):

app/scripts/proactive/templates/workflow-variables-view-template.html
amouhoub: @fviale there is an error that will be displayed anyway by the browser in the console for an 404 or 403 network call

fviale: ok

PR #985, opened 2024-08-13
Add link to Detachable window when submitting a Job

medou-boushab:


ankicabarisic (approved):
seems ok

only thing which could be improved is when using encodeURIComponent and encodeURI to ensure that no user input is injected int o these strings, as this could lead to XSS (Cross-Site Scripting) attacks.

medou-boushab:
> only thing which could be improved is when using encodeURIComponent and encodeURI to ensure that no user input is injected int o these strings, as this could lead to XSS (Cross-Site Scripting) attacks.

the browser do this by default, it's not an internal request

PR #996, opened 2025-01-10
fix-search-wf-width

oanaBiancaSchiopu:


aminelouati (approved):

app/scripts/proactive/templates/workflow-list.html
```diff
@@ -13,13 +13,13 @@
 <form class="form-inline" role="form" id="workflow-list-controls">
     <div class="form-group">
         <button type="button" class="btn create-workflow-button">Create</button>
-            <div class="input-group w-100" style="width:240px;">
-                <input type="text" id="search-workflow-input" placeholder="Workflow name" class="form-control" style="width: 200px;">
+            <div class="input-group w-100" style="width:50%;">
```
amouhoub, line 16: In this case you should remove the class w-100 because it declares width:100%

amouhoub (changes requested):

AlexFalessi (approved):

PR #1005, opened 2025-01-21
check-if-session-id-is-correct-before-login

oanaBiancaSchiopu:


app/scripts/proactive/view/LoginView.js
```diff
@@ -35,10 +35,27 @@ define(
             tryToConnect: function() {
                 if (StudioClient.isLoggedIn()) {
```
fviale, line 36: maybe rename this function to isSessionPresent()

fviale (approved):
LGTM; small comment

app/scripts/proactive/view/LoginView.js
fviale: actually I did not mean this function

I meant StudioClient.isSessionPresent()

AlexFalessi (approved):

